PR #1, opened 2018-03-06
Improved macros in Instruction.def to avoid code duplication when printing.

theo25:


dfilaretti:
From what I can see, this not only introduces "Improved macros" (as per title), but also a refactoring of the pretty printing mechanism. 

While before we had a simple structure like the following

```
void IeleInstruction::print(llvm::raw_ostream &OS, unsigned indent) const {
  std::string Indent(indent, ' ');
  OS << Indent;

  switch (getOpcode()) {
    case Ret: printRet(OS, this); break;
    case Revert: printRevert(OS, this); break;
    ...

```
we now have most of those "opcode-specific" printing functions (e.g. `printRet`) factored out into a single one (well, two actually). In particular, there's now a distinction between "intrinsic" and "other" opcodes, with slightly differing handling mechanisms (corresponding to functions `printOtherInst` and `printIntrinsic`). This new structure makes heavier use of the macros defined in `IeleInstruction.def`.

Given my still limited familiarity with the codebase I don't feel like giving a strong judgement. 
The change looks OK to me, although I must say it took me a while to understand what was going on, i.e. I found the initial version of the code much more readable (although I understand it contained some repetition). 

In addition to the above, an additional change is made in `IeleValue::printAsValue` in order to change the way a `IeleBlockVal` is printed. Again, I'm not fully familiar with the whole story of why exactly the change was required in the first place, so it would be useful if someone else could take a look.

dfilaretti:
Perhaps @dwightguth can also double-check this before me merge? Thank you.

libiele/IeleInstruction.cpp
```diff
@@ -420,174 +399,74 @@ static void printCallAt(llvm::raw_ostream &OS, const IeleInstruction *I) {
...
-//#include "IeleInstruction.def"
+#define HANDLE_IELE_OTHER_INST(N, OPC, TXT) \
+  case OPC: printOtherInst(OS, this); break;
+#include "IeleInstruction.def"
+
+#define HANDLE_IELE_INTRINSIC_INST(N, OPC, TXT) \
+  case OPC: printIntrinsic(OS, this); break;
+#include "IeleInstruction.def"
+
+  case CallAt: printCallAt(OS, this); break;
```
dwightguth: why is this still here?

theo25: I was missing the rest of the call instructions, but I just pushed a commit that adds them. Please take a look again.

dwightguth:
This looks good, but it would be good to get CI up and running. I can set up the job if you give me a command to run. Ideally I would like to see the following covered by such a command:

1. Testing the IELE tests you wrote that do end-to-end work already.
2. Testing that the textual IELE output by the compiler matches a reference file;.
3. Testing the solidity command line tests that currently pass.

dwightguth:
I think we can do that in the next PR though. But you should make that your next priority.

theo25:
Will do @dwightguth

PR #4, opened 2018-03-09
initial refactoring of end to end tests

dwightguth:
This should not be viewed as a complete piece of work, but it touches only the C++ tests, and I want to get it in so that it can live separate from the refactoring of the top level compiler that I am going to create next. So please review it with that in mind.


test/libsolidity/SolidityEndToEndTest.cpp
```diff
@@ -81,7 +81,7 @@ BOOST_AUTO_TEST_CASE(exp_operator_const)
 		}
 	)";
 	compileAndRun(sourceCode);
-	ABI_CHECK(callContractFunction("f()", bytes()), toBigEndian(u256(8)));
+	ABI_CHECK(callContractFunction("f()", encodeArgs()), encodeArgs(toBigEndian(u256(8))));
```
theo25: nit: why isn't this `vector<bytes>()` instead of `encodeArgs()` similarly to the rest?

dwightguth: No particular reason, just an artifact of how I did the replacement, but they're equivalent on the empty set of arguments, so I'm inclined to leave it for now

theo25: Sure no problem

PR #5, opened 2018-03-09
Variable declaration statement support

theo25:
Also fixed a bug in implementation of solidity's transfer/send which was uncovered by removing the by-default deposit function generation.


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -317,7 +324,39 @@ bool IeleCompiler::visit(const Break &breakStatement) {
...
+    auto const &assignments =
+      variableDeclarationStatement.annotation().assignments;
+    assert(assignments.size() == RHSValues.size() &&
+           "IeleCompiler: Missing assignment in variable declaration "
+           "statement");
+    for (unsigned i = 0; i < assignments.size(); ++i) {
+      const VariableDeclaration *varDecl = assignments[i];
+      if (varDecl) {
+        // Visit LHS. We lookup the LHS name in the function's symbol table,
+        // where we should find it.
```
dfilaretti: "where we should find it": because of the following (line 77)??

```
+  // Visit local variables.
+  for (const VariableDeclaration *local: function.localVariables()) {
+    if (local->isLocalOrReturn())
+      iele::IeleLocalVariable::Create(&Context, local->name(),
+                                      CompilingFunction);
+  }
```

theo25:
Please don't merge yet. I need to push one more fix.

theo25:
This is ready for review again. I would like to merge it after #6

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -70,10 +70,14 @@ bool IeleCompiler::visit(FunctionDefinition const& function) {
     iele::IeleFunction::Create(&Context, function.isPartOfExternalInterface(),
                                FunctionName, CompilingContract);
 
-  // Visit formal arguments.
-  for (ASTPointer<VariableDeclaration const> const& arg : function.parameters())
+  // Visit formal arguments and return parameters.
+  for (const ASTPointer<const VariableDeclaration> &arg : function.parameters())
     iele::IeleArgument::Create(&Context, arg->name(), CompilingFunction);
 
+  for (const ASTPointer<const VariableDeclaration> &ret :
```
dfilaretti: OK this makes sense (in line with explanation @theo25 gave me on Slack)

dfilaretti:
I looked at the latest change, which looks OK. 
I approved and merged #6 before this, as requested.

PR #6, opened 2018-03-09
CI support

theo25:
This PR contains a script for CI, called ciscript, as well a few manually set-up end-to-end tests with the simple auction contract.

@dwightguth please let me know if you need anything more to set up CI.


ciscript
```diff
@@ -0,0 +1,21 @@
+# clean
```
dfilaretti: Looks good. Maybe shall we mention it in the main README so people can find it easily?

dfilaretti:
I wrote a comment before, but I can't see it anymore so I'll rewrite it here. 

The CI commit looks OK. I would suggest to mention those new commands in one of the main README files for ease of use. 

Bugfix looks fine too. If I understand correctly there is only one call to `createAccountCall`; the type  of one of its args has been changed, and the (only) call has been updated accordingly. This is in line with a previous change (few days ago) where the default `deposit` function has been removed.

PR #8, opened 2018-03-12
refactor top level of compiler

dwightguth:
This code makes --iele the default, removes the --iele option, and refactors everything so that --asm and --bin work as expected to integrate more tightly with the solidity code base. In order to do this, we comment out some code for pathways we don't support yet or don't intend to support, like the gas calculation mechanism and the JSON output format.

It also fixes the CI infrastructure to correctly test for regressions.


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dfilaretti:
I took a high-level look at this commits. 

Is it true that after this refactoring it will not be possible to compile to evm? i.e. the source files that do evm compilation are still there but the functionality is not exposed anymore via the command line interface? 

Is the plan to eventually remove evm features completely, or to keep them alive alongside iele?

dwightguth:
The plan is for this fork of solidity to compile only to IELE. We will still maintain the EVM code as much as possible in order to facilitate merging, but it won't be accessible via the main code paths. The mainline solidity compiler will be used if the user wants to compile to EVM.

PR #9, opened 2018-03-12
coinbase is now implemented as iele.beneficiary

theo25:


dwightguth:
I'm going to merge this because it looks okay and it will still be a while before we get the CI sorted out.

PR #28, opened 2018-03-13
Function Modifiers support

dfilaretti:
This adds support for Function Modifiers. 

Notice however that there's still the problem of generating fresh identifiers that we discussed before. Since this looks like a more general problem that could affect other areas (e.g. function call?), I'm still submitting this for review, while we discuss how to deal with that problem separately 

Sent more detailed explanation and examples on Slack channel.


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -10,9 +10,25 @@
 using namespace dev;
 using namespace dev::solidity;
 
+// lookup a ModifierDefinition by name (borrowed from CompilerContext.cpp)
+ModifierDefinition const& IeleCompiler::functionModifier(std::string const& _name) const
```
theo25: style: use const modifier before the type, as in `const ModifierDefinition &`
style: keep `&` and `*` attached to the following word, as in `&IeleCompiler`

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -10,9 +10,25 @@
...
+{
```
theo25: style: keep `{` inline with the header of the block, it being an function signature, if/while/for header etc.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -10,9 +10,25 @@
...
+	//solAssert(!m_inheritanceHierarchy.empty(), "No inheritance hierarchy set.");
+	//for (ContractDefinition const* contract: m_inheritanceHierarchy)
+    assert(m_currentContract && "m_currentContract not set.");
```
theo25: style: Prepend the assertion string with `IeleCompiler: `

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -89,8 +105,10 @@ bool IeleCompiler::visit(FunctionDefinition const& function) {
   CompilingBlock =
     iele::IeleBlock::Create(&Context, "entry", CompilingFunction);
 
-  // Visit function's body.
-  function.body().accept(*this);
+  // Visit function body (inc modifiers). 
+	m_currentFunction = &function;
```
theo25: style: Please never use tabs! (you can set your editor to insert actual spaces when you press the tab key, if you are used to it)

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -185,6 +203,93 @@ bool IeleCompiler::visit(const Return &returnStatement) {
...
+      // Retrieve modifier definition from its name
+			ModifierDefinition const& modifier = functionModifier(modifierInvocation->name()->name());
+      
+      // Visit the modifier's parameters
+      for (const ASTPointer<const VariableDeclaration> &arg : modifier.parameters())
+        iele::IeleLocalVariable::Create(&Context, arg->name(), CompilingFunction);
+
+      // Visit the modifier's local variables
+      for (const VariableDeclaration *local: modifier.localVariables()) {
+        if (local->isLocalOrReturn())
```
theo25: See how I have implemented a similar traversal of locals/parameters/and return parameters currently in the FunctionDefinition visitor. Specifically, the way you do it here, you are skipping return parameters. I was doing the same mistake but I have a fixed version pushed in sol2iele, that you should also mimic.

libsolidity/codegen/IeleCompiler.h
```diff
@@ -70,6 +70,20 @@ class IeleCompiler : public ASTConstVisitor {
...
   bool CompilingLValue;
 
+  // Infrastructure for handling modifiers (borrowed from ContractCompiler.cpp)
+  // Lookup function modifier by name
+	ModifierDefinition const& functionModifier(std::string const& _name) const;
+  // Appends one layer of function modifier code of the current function, or the function
+	// body itself if the last modifier was reached.
+	void appendModifierOrFunctionCode();
+  unsigned m_modifierDepth = 0;
+	FunctionDefinition const* m_currentFunction = nullptr;
```
theo25: style: Rename `m_current*` to `Current*` to be uniform with the rest of the field names. Same for `m_modifierDepth`

libsolidity/codegen/IeleCompiler.cpp
dfilaretti: I tried adding the following (taken from your own function declaration visitor) but it looks like ModifierDefinition has no return parameters...! 

```
// Visit the modifier's return parameters
for (const ASTPointer<const VariableDeclaration> &ret : modifier.returnParameters())
  iele::IeleLocalVariable::Create(&Context, ret->name(), CompilingFunction);
```

I took a brief look at AST.h definitions and couldn't find it... Will have to double check.

dfilaretti:
Thanks for your comments @theo25. I have tried to apply the style adjustments you suggested, I hope it's looking better now. Please let me know if there's any further adjustment needed. 

Regarding the comment about visiting the modifiers's arguments, locals, and return parameters, that's right I actually forgot to consider the return parameters; so I tried adding that, but it seems that modifiers do not have that field. I'll leave it for now, but will take a closer look later on.

dfilaretti:
Also  I was wondering, did you see my comment about moving the instruction to retrieve the Symbol Table outside of the loop? 

I noticed this in several places:

```
for (unsigned i = 0; i < modifier.parameters().size(); ++i) {
  ...
  iele::IeleValueSymbolTable *ST = CompilingFunction->getIeleValueSymbolTable();
  ...
}
```

Is there any reason to not do it outside of the loop?

libsolidity/codegen/IeleCompiler.cpp
theo25: OK if modifiers don't have return params, feel free to remove the loop traversing them.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -203,41 +202,41 @@ bool IeleCompiler::visit(const Return &returnStatement) {
...
         iele::IeleLocalVariable::Create(&Context, arg->name(), CompilingFunction);
 
+      // TODO: double check, and delete once sure it's not needed
+      // Visit the modifier's return parameters
+      // for (const ASTPointer<const VariableDeclaration> &ret : modifier.returnParameters())
+      //   iele::IeleLocalVariable::Create(&Context, ret->name(), CompilingFunction);
+
       // Visit the modifier's local variables
       for (const VariableDeclaration *local: modifier.localVariables()) {
         if (local->isLocalOrReturn())
```
theo25: this condition is not needed anymore

theo25:
Thanks for addressing the style! I left one more minor inline comment. Regarding moving symbol table lookup out of loops: they can go out of loop, feel free to add a commit that moves them out.

theo25:
This can me merged, after you address the minor final comments

PR #30, opened 2018-03-13
Fix CI

dwightguth:
Since my other PR is blocked on @bmmoore reviewing runtimeverification/blockchain-k-plugin#7, I broke off just the parts of that PR that deal with CI so that we can get the CI target passing and unblock the other outstanding PRs on this repository.

Don't review until I get the tests passing.


dwightguth:
Closing. This also turned out to be blocked on the same PR, so there's no point keeping it separate.

PR #32, opened 2018-03-15
addmod, mulmod, conditionals, parentheses, testharness, bug fixes

### LINKED ISSUE #20: Solidity feature: Conditionals
E.g. `x ? y : z`

dwightguth:
Fixes #20 

Contains:
  * support for addmod and mulmod instructions
  * support for conditional expressions
  * support for TupleExpressions with one component (ie, parentheses)
  * fixes some bugs in the compiler in preparation for running end to end tests
  * add an assertion to fail the test on local variables of function type
  * improved test harness


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -380,7 +386,51 @@ bool IeleCompiler::visit(const InlineAssembly &inlineAssembly) {
...
+  iele::IeleValue *ConditionValue = compileExpression(condition.condition());
+  assert(ConditionValue && "IeleCompiler: failed to compile conditional condition.");
+
+  // The condition target block is the if-true block.
+  iele::IeleBlock *CondTargetBlock =
+    iele::IeleBlock::Create(&Context, "if.true");
+
+  // Connect the condition block with a conditional jump to the condition target
+  // block.
+  connectWithConditionalJump(ConditionValue, ConditionBlock, CondTargetBlock);
```
theo25: Here you should change `ConditionBlock` to `CompilingBlock` for the case where the previously called `compileExpression(condition.condition())` contained another conditional and as a result has changed the current compiling block. See the final fixed implementation for the if statement (tips of the if-fix branch).

theo25: Also I think, the way you have implemented it (e.g jump to if-true block or fall through for if-false) you need to invert the condition.

dwightguth: no? that's not what IfStatement does. It does that only if there is no false branch of the if statement, but conditional expressions always have both a true and false branch.

Also the end to end tests pass right now, which they certainly wouldn't if this were backwards.

theo25: sorry my bad, you are right

PR #38, opened 2018-03-16
Initial support for Arrays and Structs, bug fixes, minor features

theo25:
This PR adds the following:

- Support for arrays and structs, where the low level behavior of allocating and accessing structs and arrays in memory or storage has been abstracted away as runtime calls (to be implemented in next PR)
- Fixed bug in VariableDeclarationStatement
- Fixed bug in MemberAccess
- Added support for calls to `<address>.balance`
- Added support for state variable initialization in the `@init` function
- Improved IeleInstruction data structures for call instructions


theo25:
@dfilaretti I have renamed `CurrentContract` and `CurrentFunction` members of `IeleCompiler` to `CompilingContractASTNode` and `CompilingFunctionASTNode` to resolve a merge conflict between this PR and your modifiers PR.

theo25:
The 6 regression in the excluded tests are due to this PR exposing the fact we are not supporting mappings yet (previously ignored).
Also this PR removes 43 tests from the failing tests.

dwightguth:
I count 7 tests added to the excluded tests list, can you please confirm whether the seventh has the same cause (ie unsupported mappings)?

theo25:
This is weird. The extra test (test_3b34f440c30341b0d5eeb0863c3741ddaf7ce73e9ce5ca024dd47c85c040360d.sol) appears in both the added and the removed tests from the excluded list, which means it was simply reordered. This is a test that fails because of a modifier on a constructor, which makes me think that @dfilaretti maybe did not add this in its sorted place when he re-added it in his last PR?

dwightguth:
Looks like you're correct. Okay, I'm satisfied by the test coverage. I haven't looked at the code yet; I can do that after lunch if you want

theo25:
Sure

dwightguth:
I see this PR adds some new assertions. I would prefer we first merge the PR I am about to make, and then you resolve the conflicts, because my PR is going to conflict with pretty much every PR we make until it's merged, so it's best to do it ASAP.

theo25:
OK sounds good

dfilaretti:
I took a look at this, and I couldn't see anything obviously wrong or suspicious (although I must say this introduces quite a big number of changes, so I may have missed some detail). Given that it has been  tested successfully and that @dwightguth also looked at the code previously, I'll approve so that it can be merged.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -456,13 +474,29 @@ bool IeleCompiler::visit(
...
+        // Visit LHS. We lookup the LHS name in the function's symbol table,
+        // where we should find it.
         iele::IeleValue *LHSValue = ST->lookup(varDecl->name());
         solAssert(LHSValue, "IeleCompiler: Failed to compile LHS of variable "
                            "declaration statement");
+        // Check if we need to do a memory to storage copy.
+        TypePointer LHSType = varDecl->annotation().type;
+        TypePointer RHSType = RHSTypes[i];
+        iele::IeleValue *RHSValue = RHSValues[i];
+        solAssert(!shouldCopyStorageToStorage(LHSValue, RHSType),
```
dfilaretti: Maybe you could print different error messages for the two assertions here. It could be useful for debugging. Otherwise, one sees the error messages and then has to investigate further to figure out which of the two assertion triggered it.

theo25: fixed!

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -58,11 +59,14 @@ void IeleCompiler::compileContract(
   if (const FunctionDefinition *constructor = contract.constructor())
     constructor->accept(*this);
   else {
-    iele::IeleFunction *Constructor =
-      iele::IeleFunction::Create(&Context, false, "init", CompilingContract);
-    iele::IeleBlock *ConstructorBlock =
-      iele::IeleBlock::Create(&Context, "entry", Constructor);
-    iele::IeleInstruction::CreateRetVoid(ConstructorBlock);
+    CompilingFunction =
+      iele::IeleFunction::Create(&Context, true, "init", CompilingContract);
```
dwightguth: I didn't spot this in code review but this change introduced a regression in the end to end tests, because constructors in IELE are **not** public.

Issue #34, opened 2018-03-16
Allow compiler to use fresh variable names for parameters, local vars etc.


dwightguth:
Cf modifier parameter with same name as function parameter.

PR #40, opened 2018-03-19
Usage of fresh variable names in compiler & "omitted function parameter names"

dfilaretti:
Fixes #40 and #34.

This pull request makes sure that we use fresh variable names in IELE functions, so that possible clashes are avoided. 

For now, this is simply achieved by adding a unique integer suffix to user-defined variables. So for example `x` becomes `x_0` and so on. 

In addition to generating the names, we need to keep track of those mappings, so that, when visiting e.g. `x` in the user's program, the correct "generated unique name" can be found. Notice that the presence of function modifiers requires this mapping to be "modifier aware" in the sense that e.g. variable `x` can be mapped to `x_1` in a modifier and `x_123` in another. The used data structure reflect this. 

Notice that, as a side-effect, this also implements the "Omitted Function Parameter Names" solidity feature (http://solidity.readthedocs.io/en/latest/control-structures.html?highlight=anonymous#omitted-function-parameter-names); i.e. such "omitted function parameter name" also gets a unique suffix (e.g. `_324`) and then gets ignored, as per docs/spec. This is the reason of the additional tests that are now passing.


dfilaretti:
Hmm same commit appears twice... Sorry, I think I messed up with rebasing etc. Still learning my way to use git properly.

theo25:
Please add the keyword Fixes #<issue number> in the PR description, so that the PR can be linked with the appropriate issues.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1000,6 +1025,7 @@ void IeleCompiler::endVisit(const Identifier &identifier) {
   ST = CompilingContract->getIeleValueSymbolTable();
   solAssert(ST,
          "IeleCompiler: failed to access compiling contract's symbol table.");
+  // if (iele::IeleValue *Identifier = ST->lookup(VarNameMap[ModifierDepth][name])) {
```
theo25: nit: remove commented out code. I think the commented out code is incorrect here, so can be deleted safely

Issue #21, opened 2018-03-12
Solidity feature: Compound Assignment


theo25:
E.g. `x += 1`

PR #44, opened 2018-03-19
Compound assignment and post/pre increment/decrement operands

### LINKED ISSUE #43: Solidity feature: Post/Pre increment/decrement operations
e.g. `++x`, `y--`

theo25:
This PR adds support for compound assignment and post/pre increment/decrement operations.

Fixes #43 #21


libiele/IeleInstruction.cpp
```diff
@@ -379,6 +379,32 @@ IeleInstruction *IeleInstruction::CreateSStore(
...
+  solAssert(Result, "CreateLoad: Invalid lvalues");
+  solAssert(AddressValue, "CreateLoad: Invalid operands");
+
+  IeleInstruction *LoadInst = new IeleInstruction(Load, InsertAtEnd);
+  LoadInst->getIeleLValueList().push_back(Result);
+  LoadInst->getIeleOperandList().push_back(AddressValue);
+
+  return LoadInst;
+}
+
```
dfilaretti: I noticed before that those two look very similar except for the `InsertAtEnd`/`InsertBefore` parameter, which in the first case is an instruction, and in the second a block. The same goes on for `IeleInstruction(...)`, with one version accepting an instruction and one accepting a block. At closer inspection, it looks like the version of `IeleInstruction(...)` that accepts a single instruction is then going to try to retrieve the block to which the instruction belongs. 

This is perfectly fine, but I was just wondering whether it could be useful/appropriate to merge those pairs of methods so that they all directly get a block (and the before/after distinction could be handled via a flag). Just a thought so please feel free to ignore if not relevant :)

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -670,10 +681,39 @@ bool IeleCompiler::visit(const UnaryOperation &unaryOperation) {
...
+      (UnOperator == Token::Inc) ? Token::Add : Token::Sub;
+    // Compile subexpression as an lvalue.
+    iele::IeleValue *SubExprValue =
+      compileLValue(unaryOperation.subExpression());
+    solAssert(SubExprValue, "IeleCompiler: Failed to compile operand.");
+    // Get the current subexpression value, and save it in case of a postfix
+    // oparation.
+    iele::IeleValue *Before = appendLValueDereference(SubExprValue);
+    iele::IeleLocalVariable *Result = nullptr;
+    if (!unaryOperation.isPrefixOperation()) {
```
dfilaretti: I was wondering if this condition can actually be moved a few lines below, so that it can be put together with its "other half" at line 707? So instead of having two conditions testing for `unaryOperation.isPrefixOperation()` and `!unaryOperation.isPrefixOperation()`, we could have it in a single `unaryOperation.isPrefixOperation() then __ else __`.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -670,10 +681,39 @@ bool IeleCompiler::visit(const UnaryOperation &unaryOperation) {
...
+        iele::IeleLocalVariable::Create(&Context, "tmp", CompilingFunction);
+      iele::IeleInstruction::CreateAssign(Result, Before, CompilingBlock);
+    }
+
+    // Generate code for the inc/dec operation.
+    iele::IeleIntConstant *One = iele::IeleIntConstant::getOne(&Context);
+    iele::IeleLocalVariable *After =
+      appendBinaryOperator(BinOperator, Before, One);
+    // In case of a prefix operation, this is the result.
+    if (unaryOperation.isPrefixOperation())
```
dfilaretti: See previous comment at line 696

libsolidity/codegen/IeleCompiler.cpp
theo25: Fixed

libiele/IeleInstruction.cpp
theo25: If we wanted to merge them, then we can't ignore the instruction argument. We need that one because we need to know where to insert the new instruction when the flag tells us that we don't want to insert at the end. We could drop the block argument and keep one API with an instruction argument and flag telling us whether we should insert before the given instruction or at the end of the block of the given instruction. But that is not an intuitive API: in case of insertion at the end, user needs to provide an instruction instead of the block, and then the implementation needs to get back the block from the provided instruction.

PR #47, opened 2018-03-21
Make CI run all C++ tests and fix end to end test harness

### LINKED ISSUE #37: Need to run all C++ tests on every CI build of solidity
The Solidity repo has a bunch of tests already. We need to adapt the CI to run all of them and fail the build if we have any regressions, but not if tests that haven't been supported yet are failing. We've done this for the command line tests, but not for the tests written using Boost unit testing. We also need to integrate theses numbers into the summary of tests passing and failing (cf. #36)

dwightguth:
Fixes the C++ tests run via `soltest` and adds them to the set of tests run on every PR, with annotations on the tests indicating the ones currently failing.

Fixes #37


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Note that the single test added to the failing list isn't actually a test; it's a raw string in a C++ file picked up by the isolate_tests script which is not in fact a solidity program. We will need to come up with a more permanent solution eventually but for now I simply added it to the exclusion list.

ciscript
```diff
@@ -21,8 +21,13 @@ for f in $(ls iele-example-tests/auction/*.json); do
...
 
-kill -int %1
-
 # run command line tests
 ./test/ieleCmdlineTests.sh
 
+iele-test-client build/ipcfile $PORT &
+sleep 3
+
+#lrun boost tests
```
theo25: nit: `#lrun` -> `# run`

test/ExecutionFramework.cpp
```diff
@@ -110,21 +110,23 @@ void ExecutionFramework::sendMessage(std::vector<bytes> const& _arguments, std::
 	{
 	        d.to = dev::toString(m_contractAddress);
 	        BOOST_REQUIRE(m_rpc.eth_getCode(d.to, "latest").size() > 2);
+		vector<string> const& outputs = m_rpc.iele_call(d);
```
theo25: I am not sure I understand what this is doing.

dwightguth: So, the transaction receipt for a transaction in Ethereum doesn't include the return value of the function that was called by the transaction for some reason. Because of that, I modified our test harness to provide that information, but that introduced an incompatibility with the rpc structure provided by ethereum. When I found out that iohk was working on supporting that api, I had to modify the test harness back into compatibility with those rpc endpoints, which meant removing the output field from the transaction receipt. So in order to test that value, we had to go back to how we were getting that output in mainline solidity, which in this case was adapted slightly into the iele_call endpoint because iele returns a tuple of values instead of a single value. Hence this change.

theo25: OK sounds good. Thanks!

PR #48, opened 2018-03-21
Add support for internal call.

### LINKED ISSUE #18: Solidity feature: Internal Call

theo25:
This PR adds 3 regressions, but all three programs were compiling incorrectly before. The programs use function names as operands, so the iele assembler rejects them. Before the solidity compiler would incorrectly compile these names as local iele registers (so for a function named `@f` it would incorrectly compile the operand as `%f`) which made the result to seem ok. Now this has been fixed by correctly populating the symbol table of the contract with all the function names beforehand and the "correct" compilation exposes our inability to deal with function pointers. I will add an issue for that.

fixes #18


libiele/IeleFunction.h
```diff
@@ -58,12 +60,28 @@ class IeleFunction :
   static IeleFunction *Create(IeleContext *Ctx, bool isPublic,
                               const llvm::Twine &Name = "",
                               IeleContract *C = nullptr) {
-    return new IeleFunction(Ctx, isPublic, Name, C);
+    return new IeleFunction(Ctx, isPublic, false, false, Name, C);
+  }
+
+  static IeleFunction *CreateInit(IeleContext *Ctx, bool isPublic,
```
dwightguth: CreateInit should not have a parameter isPublic because it should always be false.

libiele/IeleValue.cpp
```diff
@@ -135,9 +135,16 @@ void IeleValue::printAsValue(llvm::raw_ostream &OS) const {
     OS << "%" << getName();
     break;
   case IeleGlobalVariableVal:
-  case IeleFunctionVal:
     OS << "@" << getName();
     break;
+  case IeleFunctionVal: {
```
dwightguth: This code seems to be written the same way in multiple places. Is it worth extracting into a function?

theo25:
The 2 regressions in the C++ tests are due to the following:
- 1 regression due to function pointers
- 1 regression due to not handling return params. The test was failing to compile before (hence 1 assertion failed) but now it compiles but doesn't return correct results (hence 5 failing assertions).

libiele/IeleValue.cpp
theo25: I took your advice and refactored this code

Issue #42, opened 2018-03-19
Solidity feature: 'returns'


dfilaretti:
See: http://solidity.readthedocs.io/en/v0.4.21/control-structures.html#output-parameters

PR #50, opened 2018-03-22
Tuples and implicit/omitted returns

### LINKED ISSUE #45: Solidity feature: tuple expression

dfilaretti:
This PR adds the remaining bits necessary to have tuples working and extends the "default return" mechanism so that instead of always getting `ret void` we now get the correct instruction, according to the function signature. 

Fixes #45 and #42.


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -173,9 +182,28 @@ bool IeleCompiler::visit(const FunctionDefinition &function) {
...
+        // Find Symbol Table for this function
+        iele::IeleValueSymbolTable *ST =
+          CompilingFunction->getIeleValueSymbolTable();
+        solAssert(ST,
+                  "IeleCompiler: failed to access compiling function's symbol "
+                  "table.");
+
+        // Prepare arguments for the `ret` instruction by fetching the param names
+        for (const std::string paramName : ReturnParameterNames)
+          Returns.push_back(ST->lookup(paramName));
```
theo25: Please add a `solAssert` for the result of the lookup, since we expect it will never fail.

dfilaretti: done!

dfilaretti:
The regressions are: `conditional_expression_tuples`, `bool_conversion`, `return_does_not_skip_modifier` and `no_nonpayable_circumvention_by_modifier`. 

Prior to this PR those tests were failing at an early stage because of the lack of support for tuples and partial support of returns. 

Now that they are allowed to make progress, they expose some interesting corner cases that we should definitely investigate. In other words, those are genuine failures that shall be addressed. Will discuss this in more detail on Slack and open bugs/issues.

PR #52, opened 2018-03-23
Runtime for arrays and structs 

### LINKED ISSUE #10: Solidity feature: Structs

### LINKED ISSUE #11: Solidity feature: Arrays

theo25:
This commit introduces a simple runtime for management of array and struct allocations in the local memory and storage. This runtime is then used to:
- correctly allocate arrays and structs in storage and memory;
- correctly initialize local and state variables of array or struct type;
- correctly copy arrays/structs between storage and memory.

Main points of the allocation strategy for reference types:
- Non-scalar types with non-scalar elements are now allocated in a flat way where the non-scalar elements are allocated linearly one after the other inside the object.
- Dynamically sized elements have a capped maximum size that is allocated immediately.
- All storage allocations happen at compile-time, no need for runtime management for storage other for copying.

Fixes #10 
Fixes #11


theo25:
Regressions:
 
- test_0024bebcac0ade3b83dd1930ae67631a8b812c333bf2b4977fc9e6785cfd1257.sol
- test_0846efaa317212d98487eb945f85669336a34ad61efb233f1a0a5077d0d60d53.sol
- test_40052c83b0583bcfeb4c40f80df11099054849b4760483f27a3049e4eae478d1.sol
- test_88b2deb8433a32881ef3021bf9f74ec7ab424dff58b98ac012fd93350863473a.sol
- test_930235fe21a0ba986bd1e501459705461c999ff018a914e492cc737a58ca1f2f.sol
- test_e427241de1e9715336384ad89b23bda4f60412ed5a1f09fd676cdbb0cbde9356.sol

These tests fail because they are using array literals. Before these were compiling to wrong IELE but this PR has added an assertion to fail the compilation of this feature. I will add this feature in my next PR.

- SolidityEndToEndTest/return_string
- SolidityEndToEndTest/index_access_with_type_conversion
- ABIDecoderTest/fixed_arrays

All these tests had one failing assertion before: the contract creation transaction reverted with status code 9 (malformed contract, the problem being they were calling unimplemented runtime functions). Now the creation is successful, but the checks for correct results fail, because of incorrect handling of transferring non-scalar data through account calls. I will open an issue for that and handle it in future PR.

theo25:
@dwightguth @dfilaretti This is ready for review now.

iele-example-tests/auction/auction.iele.ref
```diff
@@ -1,17 +1,23 @@
...
-@ended = 5
+@highestBidder = 4
+
+@highestBid = 5
+
+@ended = 6
 
 define @init(%_beneficiary_0, %_biddingTime_1) {
 entry:
+  sstore 7, @ielert.storage.next
```
dwightguth: should we really generate such an expensive operation as sstore when the value is never used? I guess you could argue that's the job of the optimizer but since this is going to have a big impact on gas and the optimizer isn't going to be ready for a while, it might be worth fixing.

Moreover, should memory.next be put in storage given that memory doesn't persist beyond a single call frame?

libiele/iele-rt/iele-rt.h
```diff
@@ -0,0 +1,177 @@
+R"(
+define @ielert.memory.allocate(%size) {
+entry:
+  // get next free address
+  %ptr = sload @ielert.memory.next
```
dwightguth: see above; I don't believe this should be a storage location

libiele/iele-rt/iele-rt.h
```diff
@@ -0,0 +1,177 @@
...
+
+  // compute new next free address
+  %ptr2 = add %ptr, 2
+  %next = add %ptr2, %size
+
+  // check for out-of-memory error
+  %out.of.memory = cmp gt %next, 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```
dwightguth: do we really need this? we're going to run out of gas long long before this limit is ever reached, so it seems like a waste of gas to me.

libiele/iele-rt/iele-rt.h
```diff
@@ -0,0 +1,177 @@
...
+  br %out.of.memory, throw
+
+  // allocate
+  sstore %next, @ielert.memory.next
+
+  // initialize the header
+  store %size, %ptr // size of chunk
```
dwightguth: I'm not sure I understand what this is doing, can you explain?

libiele/iele-rt/iele-rt.h
```diff
@@ -0,0 +1,177 @@
...
+  revert -1
+}
+
+define @ielert.storage.allocate(%size) {
+entry:
+  // get next free address
+  %ptr = sload @ielert.storage.next
+
+  // compute new next free address
+  %ptr2 = add %ptr, 2
```
dwightguth: this feels like a weird solution... we are using 3 storage locations for every one value that we allocate. why does this function need to do anything other than increment ielert.storage.next and then return the previous value?

libiele/iele-rt/iele-rt.h
```diff
@@ -0,0 +1,177 @@
...
+  %ptr1 = add %ptr, 1
+  sstore 0, %ptr1    // pointer to next chunk, initially null
+
+  // return the address of the header
+  ret %ptr
+}
+
+define @ielert.memory.size(%ptr) {
+entry:
+  %size = 0
```
dwightguth: again this feels like an odd solution. We are getting the size of an object by executing a loop? why aren't we just statically determining the size from the header?

iele-example-tests/auction/auction.iele.ref
theo25: I agree with both concerns here. I will post fixes.

libiele/iele-rt/iele-rt.h
theo25: I would prefer to leave this there for now, maybe with an added comment that it should probably be removed if we want to optimize for gas. See below a more detailed reply to the whole performance concern.

libiele/iele-rt/iele-rt.h
theo25: So let me explain how the runtime works. Each time you allocate an array (or struct for that matter) with a requested size S, we allocate S+2 locations. Location 0 contains the size (S), location 1 contains the address of the (S+1)th element in case the array gets extended (initialy null), rest of the locations contain the values of the elements of the array. If the array gets extended, location 1 should be updated with the address of the next "chunk", which has exactly the same structure. So essentially an array is represented as a linked list of chunks. I understand that this is not optimal neither in terms of performance, scape, nor gas. However it's a simple working solution that is mostly orthogonal to the rest of the compiler that can be improved upon later. Note that in the case that the array is not extended (e.g. structs, memory arrays), all the looping code is reduced to simply the first iteration.

libiele/iele-rt/iele-rt.h
theo25: See comment above.

libiele/iele-rt/iele-rt.h
theo25: See comment above. The code will execute more than one iterations only in case of an extended array.

libiele/iele-rt/iele-rt.h
dwightguth: I mean, my point is this is always going to succeed, so why are we even bothering to check it?

dwightguth:
Question, can all arrays be extended or only arrays declared without a size modifier?

theo25:
Only arrays that are allocated in storage and without a size modifier can be extended.

libiele/iele-rt/iele-rt.h
dwightguth: if only arrays without a size modifier and stored in storage can be extended, can we get rid of the notion of chunks for stuff that is allocated in memory?

libiele/iele-rt/iele-rt.h
```diff
@@ -0,0 +1,177 @@
...
+  %chunk.ptr = %ptr
+
+  // loop over chunks to accumulate the size
+loop:
+  %done = cmp eq %chunk.ptr, 0
+  br %done, exit.loop
```
dwightguth: it seems to me that if we have only one chunk in an array this will return the wrong size of zero.

libiele/iele-rt/iele-rt.h
```diff
@@ -0,0 +1,177 @@
...
+
+define @ielert.storage.size(%ptr) {
+entry:
+  %size = 0
+  %chunk.ptr = %ptr
+
+  // loop over chunks to accumulate the size
+loop:
+  %done = cmp eq %chunk.ptr, 0
+  br %done, exit.loop
```
dwightguth: same here

libiele/iele-rt/iele-rt.h
```diff
@@ -0,0 +1,177 @@
...
+  %found = cmp lt %index.chunk, %size.chunk
+  br %found, chunk.found
+
+  // 2. if it's not, move to the next one
+  %ptr.chunk1 = add %ptr.chunk, 1
+  %ptr.chunk = load %ptr.chunk1
+
+  // 2.1 check for out of bounds
+  %out.of.bounds = iszero %ptr.chunk
+  br %out.of.bounds, throw
```
dwightguth: is the behavior of solidity on out of bounds array access to revert the call frame? that's what you're doing here I just want to make sure you've checked that that's the correct behavior.

theo25: yes it supposed to throw: https://solidity.readthedocs.io/en/v0.3.5/control-structures.html#exceptions

Now, I don't know what status code if any should be thrown, I picked `-1` mostly arbitrarily.

libiele/iele-rt/iele-rt.h
theo25: Yes I can do that

libiele/iele-rt/iele-rt.h
theo25: Yes, I think you are right for the case an array starts at address zero (e.g. `%ptr == 0`). Will fix, good catch!

libiele/iele-rt/iele-rt.h
theo25: This is safe (unlike in the case for memory) since storage address 0 is reserved for the next free pointer.

dfilaretti:
I started looking at this at a semi high level. I'll resume tomorrow and go in more depth. From what I see, there's a library of IELE functions for memory management (`iele-rt.h`) which will be included in the contracts that require array/struct functionality. This "library" deals with things such as allocating memory for data structures, getting their size etc., and its functions are referenced from other helper functions/wrappers in `IeleCompiler` (such as e.g. `IeleCompiler::appendIeleRuntimeStorageSize` etc.). Those, in turn, make up the infrastructure to handle array and structs (e.g. as used in `doArrayAllocation` etc.). 

~~By the way, what exactly is a "runtime" in this context? Is it this the content of `iele-rt.h`? Or the overall machinery for arrays/structs?~~

~~And why is that functionality implemented in IELE? Can't it be done by the compiler? (sorry if obvious questions)~~

I'm not fully familiar with the details yet, but if I can give a general suggestion (for all of us, starting from me!) would probably be to try to add more "explanations" of the overall architecture or idea when needed, such as the one above where you explained how for an array of length S we allocate S+2 locations, and how/why we use those 2 extra locations etc. That was really good, and reading it gave me a much better feel of what's going on while reading the code. It took me 3 mins to read, but learning that same thing by going through the code only would have taken much more. I'm saying this because I find it quite challenging to go through such big changesets without having some sort of high level overview to guide me through (also, those notes could be useful for future reference too, perhaps). I understand that things become easier the more we become familiar with a codebase and become experienced with it, but well, just a though... 

I'll follow up with more detailed comments.

dfilaretti:
So I've looked into this in more detail and I think it kinda makes sense to me (i.e. I've been able to understand what the code in `iele-rt.h` does in detail and build a mental picture of how this is all used in the compiler).

I have to agree with @dwightguth on the fact that `doArrayAllocation` (& struct) is quite hard to follow. In particular, it seems to me that it is handling many cases such as 

* creation of new array (with fixed size) 
* creation of new array (dynamic size)
* copy of array  

Would it help trying to split that big function into a couple of smaller ones to handle specific cases (e.g. a function to init an empty new array and one for array copy)? I understand it may introduce some duplication though, so I'm not 100% sure this is the best way to go. 

And yes, seeing an example of what gets generated by the compiler would certainly help clarify.

iele-example-tests/auction/auction.iele.ref
theo25: Both should be fixed now

libiele/iele-rt/iele-rt.h
theo25: I removed it. I guess its good enough that we have it mentioned here, doesn't need to be in the code too.

libiele/iele-rt/iele-rt.h
theo25: I fixed this too, since now storage address 0 is not reserved anymore (the pointer to the next free location is the last state variable).

theo25:
@dwightguth @dfilaretti This is ready for review again. I have added some more detailed comments on the array/struct allocation code gen. Feel free to ask me for more details here or on slack if you need.

Also here is an example for array/struct allocation. The following solidity code
```
contract Test {
  struct S { uint x; uint y; uint[2] z; }

  S[5] data;

  function testCopyRead() returns (uint a, uint b, uint c, uint d) {
    data[2].x = 1;
    data[2].y = 2;
    data[2].z[0] = 3;
    data[2].z[1] = 4;
    S memory s = data[2];
    a = s.x;
    b = s.y;
    c = s.z[0];
    d = s.z[1];
  }
}
```
becomes
```
contract Test {

@data = 0

@ielert.storage.next = 1

define public @"testCopyRead()"() {
entry:
  %a_0 = 0
  %b_1 = 0
  %c_2 = 0
  %d_3 = 0
  %tmp = call @ielert.memory.allocate(3)
  %tmp1 = call @ielert.memory.address(%tmp, 0)
  store 0, %tmp1
  %tmp2 = call @ielert.memory.address(%tmp, 1)
  store 0, %tmp2
  %tmp3 = call @ielert.memory.allocate(2)
  %tmp4 = 0

alloc.loop:
  %tmp5 = cmp eq %tmp4, 2
  br %tmp5, alloc.exit
  %tmp6 = call @ielert.memory.address(%tmp3, %tmp4)
  store 0, %tmp6
  %tmp4 = add %tmp4, 1
  br alloc.loop

alloc.exit:
  %tmp7 = call @ielert.memory.address(%tmp, 2)
  store %tmp3, %tmp7
  %s_4 = %tmp
  %data.val = sload @data
  %tmp8 = call @ielert.storage.load(%data.val, 2)
  %tmp9 = call @ielert.storage.address(%tmp8, 0)
  sstore 1, %tmp9
  %data.val10 = sload @data
  %tmp11 = call @ielert.storage.load(%data.val10, 2)
  %tmp12 = call @ielert.storage.address(%tmp11, 1)
  sstore 2, %tmp12
  %data.val13 = sload @data
  %tmp14 = call @ielert.storage.load(%data.val13, 2)
  %tmp15 = call @ielert.storage.load(%tmp14, 2)
  %tmp16 = call @ielert.storage.address(%tmp15, 0)
  sstore 3, %tmp16
  %data.val17 = sload @data
  %tmp18 = call @ielert.storage.load(%data.val17, 2)
  %tmp19 = call @ielert.storage.load(%tmp18, 2)
  %tmp20 = call @ielert.storage.address(%tmp19, 1)
  sstore 4, %tmp20
  %data.val21 = sload @data
  %tmp22 = call @ielert.storage.load(%data.val21, 2)
  %tmp23 = call @ielert.memory.allocate(3)
  %tmp24 = call @ielert.storage.load(%tmp22, 0)
  %tmp25 = call @ielert.memory.address(%tmp23, 0)
  store %tmp24, %tmp25
  %tmp26 = call @ielert.storage.load(%tmp22, 1)
  %tmp27 = call @ielert.memory.address(%tmp23, 1)
  store %tmp26, %tmp27
  %tmp28 = call @ielert.storage.load(%tmp22, 2)
  %tmp29 = call @ielert.storage.size(%tmp28)
  %tmp30 = call @ielert.memory.allocate(%tmp29)
  %tmp31 = 0

alloc.loop32:
  %tmp33 = cmp eq %tmp31, %tmp29
  br %tmp33, alloc.exit36
  %tmp34 = call @ielert.storage.load(%tmp28, %tmp31)
  %tmp35 = call @ielert.memory.address(%tmp30, %tmp31)
  store %tmp34, %tmp35
  %tmp31 = add %tmp31, 1
  br alloc.loop32

alloc.exit36:
  %tmp37 = call @ielert.memory.address(%tmp23, 2)
  store %tmp30, %tmp37
  %s_4 = %tmp23
  %tmp38 = call @ielert.memory.load(%s_4, 0)
  %a_0 = %tmp38
  %tmp39 = call @ielert.memory.load(%s_4, 1)
  %b_1 = %tmp39
  %tmp40 = call @ielert.memory.load(%s_4, 2)
  %tmp41 = call @ielert.memory.load(%tmp40, 0)
  %c_2 = %tmp41
  %tmp42 = call @ielert.memory.load(%s_4, 2)
  %tmp43 = call @ielert.memory.load(%tmp42, 1)
  %d_3 = %tmp43
  ret %a_0, %b_1, %c_2, %d_3
}

define @init() {
entry:
  sstore 2, @ielert.storage.next
  %tmp = call @ielert.storage.allocate(5)
  %tmp1 = 0

alloc.loop:
  %tmp2 = cmp eq %tmp1, 5
  br %tmp2, alloc.exit13
  %tmp3 = call @ielert.storage.allocate(3)
  %tmp4 = call @ielert.storage.address(%tmp3, 0)
  sstore 0, %tmp4
  %tmp5 = call @ielert.storage.address(%tmp3, 1)
  sstore 0, %tmp5
  %tmp6 = call @ielert.storage.allocate(2)
  %tmp7 = 0

alloc.loop8:
  %tmp9 = cmp eq %tmp7, 2
  br %tmp9, alloc.exit
  %tmp10 = call @ielert.storage.address(%tmp6, %tmp7)
  sstore 0, %tmp10
  %tmp7 = add %tmp7, 1
  br alloc.loop8

alloc.exit:
  %tmp11 = call @ielert.storage.address(%tmp3, 2)
  sstore %tmp6, %tmp11
  %tmp12 = call @ielert.storage.address(%tmp, %tmp1)
  sstore %tmp3, %tmp12
  %tmp1 = add %tmp1, 1
  br alloc.loop

alloc.exit13:
  sstore %tmp, @data
  ret void
}

... runtime functions ...
```

@dfilaretti I agree about the explanations. I should have included some more in the origin PR comment. Regarding your question about splitting the `doArrayAllocation` `doStructAllocation` methods, I think that would result in code duplication of some fairly involved code, which would mean double the time of debugging/extending it. I have some higher-level wrappers (i.e. `appendArrayAllocation`, `appendArrayCopy`) to be used in the rest of the code.

dwightguth:
I approved but I think @dfilaretti  should also review since this is such a complex change.

dwightguth:
Update: I thought about it some more and concluded that we have a practical upper limit on array size because of prohibitive amounts of disk space, which means we can simplify the above logic somewhat. Essentially, without getting too in depth, we make the integer offset of each resizable array element take up 32 bytes in the keyspace of storage. So `int x[][];`, x[i][j] is at address(x) + i <<WordSize+256 + j<<WordSize

theo25:
Here is the code for the same example from above that is generated with the new allocation strategy. Note that there is one extra allocation at beginning of the `testCopyRead` function compared to what @dwightguth suggested in his example. This is because memory stored local variables are considered initialized to zeros at the entry of the function, i.e. code like `s.x = 1` at the very beginning of `testCopyRead` is legal (at least with the scoping rules of Solidity < 0.5, I think). So we have to allocate such variables in memory at the beginning of the function. This allocation is easily removed by a dead code elimination optimization pass. Also note that there are bounds checks that can also be easily removed by some constant-propagation + dead code elimination pass.

```
contract Test {

@data = 0

define public @"testCopyRead()"() {
entry:
  %callvalue = call @iele.callvalue()
  br %callvalue, throw
  %tmp = call @ielert.memory.allocate(4)
  %s_4 = %tmp
  %tmp2 = mul 2, 4
  %index.out.of.range = cmp lt %tmp2, 0
  br %index.out.of.range, throw
  %index.out.of.range = cmp ge %tmp2, 20
  br %index.out.of.range, throw
  %tmp3 = add @data, %tmp2
  %tmp4 = add %tmp3, 0
  sstore 1, %tmp4
  %tmp7 = mul 2, 4
  %index.out.of.range8 = cmp lt %tmp7, 0
  br %index.out.of.range8, throw
  %index.out.of.range8 = cmp ge %tmp7, 20
  br %index.out.of.range8, throw
  %tmp9 = add @data, %tmp7
  %tmp10 = add %tmp9, 1
  sstore 2, %tmp10
  %tmp13 = mul 2, 4
  %index.out.of.range14 = cmp lt %tmp13, 0
  br %index.out.of.range14, throw
  %index.out.of.range14 = cmp ge %tmp13, 20
  br %index.out.of.range14, throw
  %tmp15 = add @data, %tmp13
  %tmp16 = add %tmp15, 2
  %tmp18 = mul 0, 1
  %index.out.of.range19 = cmp lt %tmp18, 0
  br %index.out.of.range19, throw
  %index.out.of.range19 = cmp ge %tmp18, 2
  br %index.out.of.range19, throw
  %tmp20 = add %tmp16, %tmp18
  sstore 3, %tmp20
  %tmp23 = mul 2, 4
  %index.out.of.range24 = cmp lt %tmp23, 0
  br %index.out.of.range24, throw
  %index.out.of.range24 = cmp ge %tmp23, 20
  br %index.out.of.range24, throw
  %tmp25 = add @data, %tmp23
  %tmp26 = add %tmp25, 2
  %tmp28 = mul 1, 1
  %index.out.of.range29 = cmp lt %tmp28, 0
  br %index.out.of.range29, throw
  %index.out.of.range29 = cmp ge %tmp28, 2
  br %index.out.of.range29, throw
  %tmp30 = add %tmp26, %tmp28
  sstore 4, %tmp30
  %tmp32 = mul 2, 4
  %index.out.of.range33 = cmp lt %tmp32, 0
  br %index.out.of.range33, throw
  %index.out.of.range33 = cmp ge %tmp32, 20
  br %index.out.of.range33, throw
  %tmp34 = add @data, %tmp32
  %tmp36 = call @ielert.memory.allocate(4)
  call @ielert.storage.copy.to.memory(%tmp34, %tmp36, 4)
  %s_4 = %tmp36
  %tmp37 = add %s_4, 0
  %tmp38 = load %tmp37
  %a_0 = %tmp38
  %tmp39 = add %s_4, 1
  %tmp40 = load %tmp39
  %b_1 = %tmp40
  %tmp41 = add %s_4, 2
  %tmp43 = mul 0, 1
  %index.out.of.range44 = cmp lt %tmp43, 0
  br %index.out.of.range44, throw
  %index.out.of.range44 = cmp ge %tmp43, 2
  br %index.out.of.range44, throw
  %tmp45 = add %tmp41, %tmp43
  %tmp46 = load %tmp45
  %c_2 = %tmp46
  %tmp47 = add %s_4, 2
  %tmp49 = mul 1, 1
  %index.out.of.range50 = cmp lt %tmp49, 0
  br %index.out.of.range50, throw
  %index.out.of.range50 = cmp ge %tmp49, 2
  br %index.out.of.range50, throw
  %tmp51 = add %tmp47, %tmp49
  %tmp52 = load %tmp51
  %d_3 = %tmp52
  ret %a_0, %b_1, %c_2, %d_3

throw:
  revert -1
}

define @init() {
entry:
  ret void
}

define @ielert.memory.allocate(%size) {
entry:
  // get next free address
  // note that the address 0 of memory is reserved for a pointer to
  // the last allocated memory location
  %last.allocd = load 0
  %ptr = add %last.allocd, 1

  // compute new last allocd address
  %last.allocd = add %last.allocd, %size

  // allocate
  store %last.allocd, 0

  // return the address of the header
  ret %ptr
}

define @ielert.storage.copy.to.memory(%from, %to, %size) {
copy.loop:
  %done = iszero %size
  br %done, exit
  %data = sload %from
  store %data, %to
  %from = add %from, 1
  %to = add %to, 1
  %size = sub %size, 1
  br copy.loop

exit:
  ret void
}
... rest of runtime ...
}
```

theo25:
@dwightguth @dfilaretti This is ready for review again. See my comment above, as well as the updated PR description and regressions comment (second from the top).

dfilaretti:
The new compilation example looks good to me, although I feel I'm not as well versed in the details of this as @dwightguth is, so he should probably take a look and have the last word! 

By the way, there were some newly compiling tests containing events that are now getting stuck on arrays. Once this is done let me know if it all works or if there's something I need to investigate.

test/libsolidity/SolidityTypes.cpp
```diff
@@ -19,7 +19,7 @@
  * @date 2015
  * Unit tests for the type system of Solidity.
  */
-
+/*
```
dwightguth: I'm not sure I like the idea of tests that are still applicable to the compiler being commented out... Can we fix them so that they compile?

theo25: Fixed

PR #56, opened 2018-03-23
stage 1 of migration to make solidity to iele tests work with mantis

dwightguth:
We are restructuring the JSON RPC slightly to clean it up and match the mantis implementation.


dwightguth:
Jenkins: test this please

PR #66, opened 2018-03-26
fix byte encoding of integers

dwightguth:
positive integers whose MSB is >= 128 need an extra zero byte at the beginning of their encoding. I'm not entirely sure why this was passing before, but it was definitely wrong. This should fix that.


dwightguth:
Jenkins: test this please

PR #73, opened 2018-03-28
Removes a forgotten "not implemented yet" assertion for Solidity's assert.

### LINKED ISSUE #58: Solidity feature: assert
Quite a few contracts failed on the "non implemented yet" assertion inside `appendInvalid`. 
I wasn't quite sure about what that should do, but by looking at who calls `appendInvalid`, it seems this should handle assertions.

theo25:
Fixes #58


theo25:
This is ready for review.

PR #74, opened 2018-03-28
add super keyword

dwightguth:
supports super.foo()

Doesn't yet support directly invoking a particular base class's members by means of Foo.foo();


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1477,10 +1511,21 @@ void IeleCompiler::endVisit(const Identifier &identifier) {
   if (const MagicVariableDeclaration *magicVar =
          dynamic_cast<const MagicVariableDeclaration *>(declaration)) {
     switch (magicVar->type()->category()) {
-    case Type::Category::Contract:
+    case Type::Category::Contract: {
+      const ContractType *type = dynamic_cast<const ContractType *>(magicVar->type().get());
+      if (type->isSuper()) {
+        //should only be reached if super is not part of a member access expression
+        // in this case we evaluate to the current object since that is the same as "thia" object
```
theo25: typo: thia -> this

PR #78, opened 2018-03-30
Bounded types and type conversions

### LINKED ISSUE #14: Solidity feature: Bounded Value Types
E.g. `int64`, `bytes16` etc.

dwightguth:
Adds:
* bounded value types (address, int8-int248, uint8-uint248, bytes1-bytes32)
* type conversions on all currently supported types except tuples
* length member for fixed bytes
* indexing fixed bytes
* left and right shift

Regressions:
* most of the tests that no longer compile are compilation tests that test the fixed point arithmetic types, which solidity-to-evm does not yet support either.
* the one exception is conditional_expression_tuples, which fails because I didn't implement type conversion on tuple types.

Fixes #14


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -2074,25 +2194,181 @@ iele::IeleLocalVariable *IeleCompiler::appendBinaryOperator(
...
+    case Token::Mod:
+    case Token::SHL: appendMask(Result, Result, nbytes, issigned); break;
+    case Token::Mul:
+    case Token::Exp: break; //handled above
+    case Token::Equal:
+    case Token::NotEqual:
+    case Token::GreaterThanOrEqual:
+    case Token::LessThanOrEqual:
+    case Token::GreaterThan:
+    case Token::LessThan: // can't
```
theo25: nit : remove comment

dwightguth:
Jenkins: test this please

PR #81, opened 2018-04-02
This keyword and account calls

### LINKED ISSUE #79: this
Support the this keyword (evaluates to the address of the current contract)

dwightguth:
Does not yet support account calls of reference types.

Ready for review.

Fixes #79


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

libiele/IeleInstruction.cpp
```diff
@@ -206,23 +206,29 @@ IeleInstruction *IeleInstruction::CreateCondBr(
...
     IeleInstruction *InsertBefore) {
-  solAssert(Callee && AddressValue && TransferValue && GasValue,
-         "CreateAccountCall: Invalid operands");
-
-  IeleInstruction *AccountCallInst = new IeleInstruction(CallAt, InsertBefore);
-  if (StatusValue)
-    AccountCallInst->getIeleLValueList().push_back(StatusValue);
+  if (StaticCall) {
+    solAssert(Callee && StatusValue && AddressValue && !TransferValue && GasValue,
+              "CreateAccountCall: Invalid operands");
```
dfilaretti: Perhaps having different error messages in those two cases (this and the one above) could make debugging easier when the assertions are hit?

dwightguth: Assertion failures include the line number of the assertion, so I'm not really sure that's necessary.

libiele/IeleInstruction.cpp
```diff
@@ -232,23 +238,30 @@ IeleInstruction *IeleInstruction::CreateAccountCall(
...
     IeleLocalVariable *StatusValue,
     llvm::SmallVectorImpl<IeleLocalVariable *> &LValues,
     IeleGlobalValue *Callee, IeleValue *AddressValue,
     IeleValue *TransferValue, IeleValue *GasValue,
     llvm::SmallVectorImpl<IeleValue *> &ArgumentValues,
     IeleBlock *InsertAtEnd) {
-  solAssert(Callee && AddressValue && TransferValue && GasValue,
-         "CreateAccountCall: Invalid operands");
+  if (StaticCall) {
+    solAssert(Callee && StatusValue && AddressValue && !TransferValue && GasValue,
```
dfilaretti: What about gaving `solAssert(Callee && StatusValue && AddressValue && GasValue ...` once just before the branch, and then simply `solAssert(TransferValue ...` and `solAssert(!TransferValue ...` inside?

dwightguth: I could do that I guess, but I'm not sure I see the need. If this assertion fails, you can just set a catchpoint in gdb and inspect the values of the pointers. The benefit to ease of debugging would be pretty small. @theo25 what do you think?

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1791,6 +1878,31 @@ iele::IeleValue *IeleCompiler::compileExpression(const Expression &expression) {
...
+
+  // Visit expression.
+  CompilingLValue = false;
+  expression.accept(*this);
+
+  // Expression visitors should store the value that is the result of the
+  // compiled for the expression computation in the CompilingExpressionResult
+  // field. This helper should only be used when a scalar value (or void) is
+  // expected as the result of the corresponding expression computation.
+  solAssert(CompilingExpressionResult.size() >= numExpressions, "expression visitor did not set enough result values");
```
theo25: I think the check here should be `==` rather than ` >=`

libiele/IeleInstruction.cpp
theo25: It's fine either way, you don't have to change it.

libiele/IeleInstruction.cpp
dfilaretti: Well yes, it's certainly no big deal (as for my other comment), I just thought it would be better to avoid repeating things. It will not make a huge difference, I guess it's a matter of preference.

PR #83, opened 2018-04-03
Inheritance fixes

dwightguth:
Fixes a bug involving resolving function overrides and also implements direct member access of base class members.

Ready for review.


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1404,10 +1417,51 @@ bool IeleCompiler::visit(const MemberAccess &memberAccess) {
...
+          case FunctionType::Kind::External:
+          case FunctionType::Kind::Creation:
+          case FunctionType::Kind::DelegateCall:
+          case FunctionType::Kind::CallCode:
+          case FunctionType::Kind::Send:
+          case FunctionType::Kind::BareCall:
+          case FunctionType::Kind::BareCallCode:
+          case FunctionType::Kind::BareDelegateCall:
+          case FunctionType::Kind::Transfer:
+            break;
```
theo25: Why is break the correct behavior here? Also these cases are not properly indented.

dwightguth: good catch, that was a bug. I think I fixed it, can you re-review?

theo25: Yes this looks better. but still the `case` clauses (except the first) are not indented correctly. They should be at the same level as the the `switch`.

PR #88, opened 2018-04-05
Dynamic contract creation with new

dwightguth:
Ready for review.


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1803,7 +1809,13 @@ bool IeleCompiler::visit(const MemberAccess &memberAccess) {
     break;
   }
   case Type::Category::Function:
-    if (member == "selector")
+    if (member == "value" || member == "gas") {
+      llvm::SmallVector<iele::IeleValue*, 4> CalleeValues;
```
theo25: nit : This `4` I think should be `2`.

PR #91, opened 2018-04-06
Log instructions

dfilaretti:
Fixes #91 

This adds support for `log0`,`log1`,`log2`,`log3` and `log4`.


dfilaretti:
Can now be reviewed.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1632,8 +1632,55 @@ bool IeleCompiler::visit(const FunctionCall &functionCall) {
...
+      CompilingBlock);
+    // i.e. %next.free = add %last.used, 1
+    iele::IeleInstruction::CreateBinOp(iele::IeleInstruction::Add,
+                                       NextFree,
+                                       LastUsed, 
+                                       iele::IeleIntConstant::getOne(&Context),
+                                       CompilingBlock);
+
+    // Store the data argument in memory
+    iele::IeleInstruction::CreateStore(DataArg,
```
dwightguth: I think this should be a 32-byte width store, and you should remove the AlignRights from the tests you modified. We can't really do that for events because the data field has variable width, but with the low level log functions, whose data argument is of type bytes32, it will make the compatibility with EVM better.

theo25: I agree with @dwightguth here. Other than that I am fine with the changes so you can merge as soon as this fixed.

dfilaretti: Fixed! 

Notice this however: 

```
BOOST_AUTO_TEST_CASE(event)
{
	char const* sourceCode = R"(
		contract ClientReceipt {
			event Deposit(address indexed _from, bytes32 indexed _id, uint _value);
			function deposit(bytes32 _id, bool _manually) payable {
				if (_manually) {
					bytes32 s = 0x19dacbf83c5de6658e14cbf7bcae5c15eca2eedecf1c66fbca928e4d351bea0f;
					log3(bytes32(msg.value), s, bytes32(msg.sender), _id);
				} else {
					Deposit(msg.sender, _id, msg.value);
				}
			}
		}
	)";
	compileAndRun(sourceCode);
	u256 value(18);
	u256 id(0x1234);
	for (bool manually: {true, false})
	{
		callContractFunctionWithValue("deposit(bytes32,bool)", value, id, manually);
		BOOST_REQUIRE_EQUAL(m_logs.size(), 1);
		BOOST_CHECK_EQUAL(m_logs[0].address, m_contractAddress);
		BOOST_CHECK_EQUAL(h256(m_logs[0].data, h256::AlignRight), h256(u256(value)));
		BOOST_REQUIRE_EQUAL(m_logs[0].topics.size(), 3);
		BOOST_CHECK_EQUAL(m_logs[0].topics[0], dev::keccak256(string("Deposit(address,bytes32,uint256)")));
		BOOST_CHECK_EQUAL(m_logs[0].topics[1], h256(m_sender, h256::AlignRight));
		BOOST_CHECK_EQUAL(m_logs[0].topics[2], h256(id));
	}
}
```

Here in one branch it calls `event` and in the other it calls `log`. The check, used for both case, still keeps the `h256::AlignRight` otherwise the `event` case would break; as it is now, works for both. 

We could probably use a condition i.e. if in the `log` case, do not use `h256::AlignRight`, otherwise use it. Not sure whether we need/want it though...

PR #94, opened 2018-04-06
base constructor parameters

### LINKED ISSUE #16: Solidity feature: Inheritance
Subfeatures:

* [x] inherited members
* [x] base constructors without arguments
* [ ] base constructor arguments
* [x] super keyword
* [x] direct access to base class members.

dwightguth:
Ready for review.

Fixes #16


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -2635,14 +2633,45 @@ void IeleCompiler::appendInvalid(iele::IeleValue *Condition) {
...
+    }
+
+    for (const auto &base : def->baseContracts()) {
+      auto baseContract = dynamic_cast<const ContractDefinition *>(
+        base->name().annotation().referencedDeclaration);
+      solAssert(baseContract, "Must find base contract in inheritance specifier");
+      if (baseArguments.count(baseContract->constructor()) == 0) {
+        baseArguments[baseContract->constructor()] = &base->arguments();
+      }
+    }
```
theo25: I am little confused as to why the if statement (lines 2641-2651) for loop (lines 2653-2660) need to be here and not in the place I marked below.

dwightguth: line 2687 fires only when you find the immediate super contract of the current contract in the linearized inheritance hierarchy, but any more-derived contract can override the constructor arguments passed to a more-base contract. Thus we must choose the most-derived set of constructor arguments.

theo25: OK I thought line 2687 fires when you find the current contract itself not its immediate super contract. But I think I see your point. I will approve then.

PR #97, opened 2018-04-09
Delete, array copy, and array length members

### LINKED ISSUE #63: Solidity feature: delete
See e.g. `deleteLocals` in `SolidityEndToEndTest.cpp`

dwightguth:
This PR contains the code for the delete expression and the length member of the array type, but in the process of testing I also discovered a serious flaw in how we were performing copying of reference types, which forced me to go back to the earlier tactic of recursing over the type of the reference and generating code to perform the copy. We still fall back on the prior method of copying in the runtime in certain cases when it can still be done safely.

Regressions:
* one failing test that fails to compile because it tries to copy a struct that contains a mapping. @theo25 will need to handle this case in his PR.

Tests now compiling but not passing:
* dynamic_arrays
* dynamic_nested_arrays
* struct_containing_bytes_copy_and_delete

The first two fail because of the known issue of passing reference values between contracts, and the third fails because of the unimplemented bytes and string types.

Fixes #63 

Do not review yet.


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1126,6 +1131,161 @@ bool IeleCompiler::visit(const UnaryOperation &unaryOperation) {
...
+          CompilingBlock) ;
+
+      iele::IeleInstruction::CreateBinOp(
+          iele::IeleInstruction::Add, Element, LValue,
+          iele::IeleIntConstant::getOne(&Context),
+          CompilingBlock);
+    } else {
+      iele::IeleInstruction::CreateAssign(
+        SizeVariable, iele::IeleIntConstant::Create(&Context, bigint(arrayType.length())),
+        CompilingBlock);
```
theo25: you can avoid this assignment  by not initializing `SizeVariable` to point to a new `IeleLocalVariable` but rather to `nullptr`. Then in the `isDynamicallySized` case you allocate the local variable, and in the `else` case you simply do `SizeVariable = iele::IeleIntConstant::Create(&Context, bigint(arrayType.length()))`

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1126,6 +1131,161 @@ bool IeleCompiler::visit(const UnaryOperation &unaryOperation) {
...
+      iele::IeleInstruction::CreateAssign(
+        Element, LValue, CompilingBlock);
```
theo25: You can avoid this assignment too in a similar way to the above

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -2187,7 +2416,8 @@ bool IeleCompiler::visit(const IndexAccess &indexAccess) {
         iele::IeleInstruction::Add, AddressValue, ExprValue, OffsetValue,
         CompilingBlock);
 
-    if (CompilingLValue || !elementType.isValueType()) {
+    if (CompilingLValue || (!elementType.isValueType() && 
+        (!elementType.isDynamicallySized() || type.location() == DataLocation::Storage))) {
```
theo25: isn't this the same as
```
if (CompilingLValue || !(elementType.isDynamicallySized() && type.location() == DataLocation::Memory))
```

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -2567,6 +2798,36 @@ iele::IeleValue *IeleCompiler::compileLValue(const Expression &expression) {
   return Result;
 }
 
+void IeleCompiler::compileLValues(
+    const Expression &expression,
+    llvm::SmallVectorImpl<iele::IeleValue *> &Result) {
+  // Save current expression compilation status.
+  bool SavedCompilingLValue = CompilingLValue;
```
theo25: I am wondering if we should also save `CompilingLValueArrayElementSize` here and in `compileLValue`

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -2846,22 +3151,301 @@ iele::IeleValue *IeleCompiler::appendStructAllocation(const StructType &type) {
...
+          iele::IeleInstruction::Add, ElementFrom, From,
+          iele::IeleIntConstant::getOne(&Context),
+          CompilingBlock);
+    } else {
+      iele::IeleInstruction::CreateAssign(
+        SizeVariableFrom, iele::IeleIntConstant::Create(&Context, bigint(arrayType.length())),
+        CompilingBlock);
+      iele::IeleInstruction::CreateAssign(
+        ElementFrom, From, CompilingBlock);
+    }
```
theo25: you can avoid these assignments with a similar way as I described before

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -2846,22 +3151,301 @@ iele::IeleValue *IeleCompiler::appendStructAllocation(const StructType &type) {
...
+          iele::IeleInstruction::Add, ElementTo, To,
+          iele::IeleIntConstant::getOne(&Context),
+          CompilingBlock);
+    } else {
+      iele::IeleInstruction::CreateAssign(
+        SizeVariableTo, iele::IeleIntConstant::Create(&Context, bigint(toArrayType.length())),
+        CompilingBlock);
+      iele::IeleInstruction::CreateAssign(
+        ElementTo, To, CompilingBlock);
+    }
```
theo25: same here for avoiding the assignments

libsolidity/codegen/IeleCompiler.cpp
dwightguth: SizeVariable needs to be a register because it is also used as the loop counter below.

libsolidity/codegen/IeleCompiler.cpp
dwightguth: Similar here; element is incremented on each loop iteration, so I'd rather not reuse LValue as that would modify the lvalue passed to this function.

libsolidity/codegen/IeleCompiler.cpp
dwightguth: This is not equivalent if elementType is a value type, because isDynamicallySized() will still return false on a value type, making the expression you wrote true when we want it to be false for value types.

libsolidity/codegen/IeleCompiler.cpp
dwightguth: So this is actually a result of compiling an lvalue, like CompilingExpressionResult. If we saved and restored it in this function, it wouldn't really behave correctly. I agree what I have now is kind of hackish though... Do you have a better suggestion?

libsolidity/codegen/IeleCompiler.cpp
dwightguth: similar problem as above

libsolidity/codegen/IeleCompiler.cpp
dwightguth: similar problem as above

libsolidity/codegen/IeleCompiler.cpp
theo25: OK makes sense

libsolidity/codegen/IeleCompiler.cpp
theo25: Also agreed for this one

libsolidity/codegen/IeleCompiler.cpp
theo25: Fair enough

libsolidity/codegen/IeleCompiler.cpp
theo25: I was thinking of maybe using the `LValue` class that the evm backend is using or something similar to return all the information resulting from an lvalue compilation. But this is fine for now, since it's working I guess

PR #98, opened 2018-04-10
API for getting the bitwidth of types with fixed bitwidth representation

theo25:
This PR add a method in the `Type` class of the Solidity AST that returns the number of bits needed to represent values of the various types if they can be represented in a fixed number of bits through serialization. Else it returns 0.

The API hasn't been implemented for the `MappingType`, as it will be part of the next PR.

This is ready for review.

### DIFF
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -2452,7 +2452,7 @@ unsigned FunctionType::calldataEncodedSize(bool _padded) const
 bigint FunctionType::getFixedBitwidth() const
 {
 	if (m_kind == Kind::External || m_kind == Kind::Internal)
-		return bigint(80);
+		return bigint(160);
 	else
 		solAssert(false, "Bitwidth of non-storable function type requested.");
 }
--- a/libsolidity/ast/Types.cpp
+++ b/libsolidity/ast/Types.cpp
@@ -1420,6 +1420,15 @@ bool ArrayType::isDynamicallyEncoded() const
 	return isDynamicallySized() || baseType()->isDynamicallyEncoded();
 }
 
+bigint ArrayType::getFixedBitwidth() const {
+	if (isDynamicallySized())
+		return bigint(0);
+
+	if (length() > MAX_ARRAY_SIZE)
+		BOOST_THROW_EXCEPTION(Error(Error::Type::TypeError) << errinfo_comment("Array too large."));
+	return length() * baseType()->getFixedBitwidth();
+}
+
 bigint ArrayType::storageSize() const
 {
 	if (isDynamicallySized())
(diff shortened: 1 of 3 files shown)


libsolidity/ast/Types.cpp
```diff
@@ -2416,6 +2449,14 @@ unsigned FunctionType::calldataEncodedSize(bool _padded) const
 	return size;
 }
 
+bigint FunctionType::getFixedBitwidth() const
+{
+	if (m_kind == Kind::External || m_kind == Kind::Internal)
+		return bigint(80);
```
dwightguth: why this value? we haven't actually implemented function pointers yet.

libsolidity/ast/Types.cpp
```diff
@@ -2062,6 +2090,11 @@ string TupleType::toString(bool _short) const
 	return str + ")";
 }
 
+bigint TupleType::getFixedBitwidth() const
+{
+	solAssert(false, "Bitwidth of non-storable tuple type requested.");
```
dwightguth: are you sure we'll never hit this assertion?

libsolidity/ast/Types.h
```diff
@@ -202,23 +202,29 @@ class Type: private boost::noncopyable, public std::enable_shared_from_this<Type
...
-	/// @returns true if the type is dynamically encoded in the ABI
+	/// @returns true if the type doesn't fit in a fixed number of storage/memory locations.
 	virtual bool isDynamicallyEncoded() const { return false; }
+    /// If the values of the type can be represented as fixed-width integers, @returns
+    /// the bitwidth else returns zero.
+    virtual bigint getFixedBitwidth() const { return bigint(0); }
 	/// @returns the number of storage slots required to hold this value in storage.
 	/// For dynamically "allocated" types, it returns the size of the statically allocated head,
 	virtual bigint storageSize() const { return 1; }
+	/// @returns the size of this data type in bytes when stored in memory.
```
dwightguth: this isn't actually in bytes, is it?

libsolidity/ast/Types.h
```diff
@@ -691,6 +702,7 @@ class ContractType: public Type
 	{
 		return encodingType()->calldataEncodedSize(_padded);
 	}
+	virtual bigint getFixedBitwidth() const override { return bigint(80); }
```
dwightguth: shouldn't this be 160?

libsolidity/ast/Types.cpp
theo25: Yes I know. I just had to return something so I used the size of an address. We may need to change that in the future. Do you want me to replace with assertion?

libsolidity/ast/Types.cpp
theo25: Well, TupleType::storageSize has the same assertion so yes I am confident. Tuples are not first class citizens (only used for assignements and returning multiple values as far as I know)

libsolidity/ast/Types.h
theo25: yes it's in slots. I will fix the comment

libsolidity/ast/Types.cpp
dwightguth: I think it would be better to have a failing assertion here, yes.

PR #99, opened 2018-04-11
Encoding

### LINKED ISSUE #80: Encoding function to combine multiple integers into a single bytestring
This will be needed in #25 to store the non-indexed arguments. 
In the EVM compiler, this is done by calling `utils().abiEncode`.

dfilaretti:
Fixes #80. 
Introduces an encoding function that takes multiple arguments and encodes them as a single byte string. 
Not all possible cases are covered yet (e.g. `array`, `bytes`, `string` etc.). 

Do not review yet.


dfilaretti:
There are a bunch of tests from the `failing-tests` list that are now failing; this is due to the fact that they were sending arguments such as `bytes` to events, a case we don't support yet, but we didn't catch before.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1332,6 +1332,137 @@ bool IeleCompiler::visit(const BinaryOperation &binaryOperation) {
...
+  iele::IeleLocalVariable *ArgLen = 
+    iele::IeleLocalVariable::Create(&Context, "arg.len", CompilingFunction);
+ 
+  iele::IeleInstruction::CreateAssign(
+    CrntPos, iele::IeleIntConstant::getZero(&Context), CompilingBlock);    
+  for (unsigned i = 0; i < arguments.size(); i++) {
+    switch (types[i]->category()) {
+      case Type::Category::Bool:
+      case Type::Category::Integer: {
+        if (types[i] -> isValueType()) {
```
dwightguth: bool and integer are both value types, so you shouldn't need this if block.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1332,6 +1332,137 @@ bool IeleCompiler::visit(const BinaryOperation &binaryOperation) {
...
+              CompilingBlock);    
+            iele::IeleInstruction::CreateStore1(
+              ArgValue, NextFree, CrntPos, 
+              ArgTypeSize, CompilingBlock);
+            iele::IeleInstruction::CreateBinOp(
+              iele::IeleInstruction::Add, CrntPos, CrntPos, 
+              ArgTypeSize, CompilingBlock);
+          }
+          else { // Arbitrary precision
+            bigint argWidth    = bigint(5);  // TODO: dummy - use actual width!
```
dwightguth: obviously we need to fix this before we merge... I should have a PR for the iele-semantics ready soon which will have a log base 2 operator.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1332,6 +1332,137 @@ bool IeleCompiler::visit(const BinaryOperation &binaryOperation) {
...
+              ArgTypeSize, CompilingBlock);
+            iele::IeleInstruction::CreateStore1(
+              ArgValue, NextFree, CrntPos, ArgLen, CompilingBlock);
+            iele::IeleInstruction::CreateBinOp(
+              iele::IeleInstruction::Add, CrntPos, CrntPos, ArgLen, CompilingBlock);
+          }
+        }
+        break;
+      }
+      case Type::Category::Array: {
```
dwightguth: I'd feel more comfortable with this PR if we had more types supported. Specifically, I'd like to see Array, Struct, FixedBytes, Contract, and Enum supported. Don't worry about the case of ArrayType where isByteArray is true though...

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1643,13 +1776,15 @@ bool IeleCompiler::visit(const FunctionCall &functionCall) {
...
+        nonIndexedArgTypes.push_back(arguments[arg]->annotation().type);
+        nonIndexedParamTypes.push_back(function.parameterTypes()[arg]);
       }
     }
 
     // Since we don't have encoding of multiple values into a bytestring yet, 
     // we assume at most a single non-indexed arg for now. TODO: once we have encoding, remove this. 
-    solAssert(NonIndexedArguments.size() <= 1, 
-              "Only a single non-indexed param is allowed (temporarily)");
+    // solAssert(NonIndexedArguments.size() <= 1,
```
dwightguth: can we delete these four lines instead of commenting them out?

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1669,13 +1804,25 @@ bool IeleCompiler::visit(const FunctionCall &functionCall) {
...
+
 
     // Store non-indexed args in memory (for now, only one)
     // TODO: encode multiple non-indexed arguments into a single bytestring
-    if (NonIndexedArguments.size() > 0)
-      iele::IeleInstruction::CreateStore(NonIndexedArguments[0],
-                                        NextFree, 
-                                        CompilingBlock);
+    if (NonIndexedArguments.size() > 0) {
+      // iele::IeleValue *EncodedArgs = encoding(NonIndexedArguments, nonIndexedArgTypes, nonIndexedParamTypes, NextFree);
```
dwightguth: similarly, can we delete this commented out code?

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1669,13 +1804,25 @@ bool IeleCompiler::visit(const FunctionCall &functionCall) {
...
+      // iele::IeleInstruction::CreateStore(EncodedArgs,
+      //                                    NextFree, 
+      //                                    CompilingBlock);
+      encoding(NonIndexedArguments, nonIndexedArgTypes, nonIndexedParamTypes, NextFree);
```
dwightguth: it looks to me like the reason you have two sets of TypePointers is because of the arg and param types. What you should do instead is pass a single list of types to encoding, but iterate over the arguments to the function and call appendTypeConversion to convert from the argument type to the param type.

theo25: I would also like to add that I don't see the reason why you call the low-level encoding here and also repeat the memory allocation (even in the case where we don't have any indexed arguments. I think it's better to call the higher-level encoding that does the allocation inside this `if` and remove the allocation above alltogether.

dfilaretti:
Jenkins: test this please

libsolidity/codegen/IeleCompiler.cpp
dfilaretti: fixed

libsolidity/codegen/IeleCompiler.cpp
dfilaretti: done

libsolidity/codegen/IeleCompiler.cpp
dfilaretti: done

libsolidity/codegen/IeleCompiler.cpp
dfilaretti: @theo25 what do you mean by repeating the memory allocation? I think I'm doing it only once here: the code for `event` does the allocation, then calls `encoding` and passes it the address of the allocated cell, so it's only one allocation. 

It's true that it does the allocation even if there are no non-indexed parameters, but this is because the IELE instruction `log` always requires a memory location as its first argument (even if there are no non-indexed parameters hence nothing to store in it); and it turns out, in order for the end-to-end tests to pass, this cell has to be uninitialised (I remember discussing this with @dwightguth, as I was initially initialising that cell to zero, but the tests failed that way). 

Obviously this may not be the best solution, but that's why it's done this way right now...

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1332,6 +1332,137 @@ bool IeleCompiler::visit(const BinaryOperation &binaryOperation) {
...
+  iele::IeleLocalVariable *NextFree =
+      iele::IeleLocalVariable::Create(&Context, "next.free", CompilingFunction);
+  iele::IeleInstruction::CreateLoad(
+    LastUsed, iele::IeleIntConstant::getZero(&Context), CompilingBlock);
+  iele::IeleInstruction::CreateBinOp(
+    iele::IeleInstruction::Add, NextFree, LastUsed, 
+    iele::IeleIntConstant::getOne(&Context), CompilingBlock);
+
+  // Call version of `encoding` that writes destination loc (NextFree)
+  encoding(arguments, argTypes, paramTypes, NextFree);
```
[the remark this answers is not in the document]
dfilaretti: I don't mind renaming to `doEncoding`, but I think @dwightguth initially suggested having this second version as an overload. I have no strong opinion on this to be honest. Whatever you guys think it's better.

libsolidity/codegen/IeleCompiler.cpp
dfilaretti: @dwightguth I see what you mean about the two sets of TypePointers. I was actually missing the `appendTypeConversion`, so I am adding it now. But I'm still not entirely sure how to handle those two sets of type pointers; if I only pass one, say the argument types, I'll then need to pass the function in order to iterate through its parameter types, right? (e.g. as in `compileFunctionArguments`).

dfilaretti:
Here is a summary of the regressions. 

Need to adjust expected value to take encoding into account:

- `event`
- `event_emit`

Need to check what `encodeLogs` actual do in the test harness, and change it (similarly to the above cases):

- `events_with_same_name`
- `events_with_same_name_inherited`
- `events_with_same_name_inherited_emit`

Encoding not supported:

- `event_anonymous_with_topics`

dfilaretti:
Regarding those tests failing because of `encodeLogs`, it seems that we need to keep `encodeLogs` from `ExecutionFramework.h` in sync with our `encoding` functions from `IeleCompiler`. 
In particular, `encodeLogs` calls `encode` to encode actual values.

libsolidity/codegen/IeleCompiler.cpp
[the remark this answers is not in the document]
theo25: Overloads should in general be used for the same (or similar) functionality on different types of arguments. If the methods return different things and in general do different things I think they should be named differently, or else it's confusing

libsolidity/codegen/IeleCompiler.cpp
theo25: I mean there is duplicate code for generating the allocation code. It would be nice to at least refactor it in its own function, although I still think you can simply use the high-level encoding function here

dwightguth:
Closing in favor of #114

Issue #12, opened 2018-03-12
Solidity feature: Mappings


theo25:
- [x] mappings with integral keys with both finite and infinite keyspaces
- [x] mappings with hashed keyspaces
- [ ] mappings with dynamically sized keyspaces (e.g. strings)
- [x] delete for mappings
- [x] copying of structs that have mapping members

PR #104, opened 2018-04-11
Mappings

theo25:
Initial implementation of mappings

- Implemented storage allocation of mappings with finite and infinite keyspaces.
- Implemented index access for mappings with finite and infinite keyspaces.
- Implemented hashed keyspaces.
- Implemented delete for mappings.
- Implemented copying of structs with mapping members.

- Missing support for arrays as keys. Mainly I have seen `string` to be used as a key type. We should revisit that after we implement encoding.
- Missing support for mappings accessors. This should be dealt in its own pr along with complext accessors for arrays/structs.

#12 

This is ready for review.


theo25:
Jenkins: test this please

theo25:
Jenkins: test this please

theo25:
Jenkins: test this please

theo25:
An end to end test (struct_reference) is passing now and contains a recursive struct. I am not sure why, as I think I made recursive structs fail prematurely. I will investigate.

Update: the problem is due to problematic computation of storage size for structs with mapping members. I will push a fix that will make the test fail, as it should at this point since it contains a recursive struct.

dwightguth:
This looks fine, but are there really only two end to end tests that this PR makes pass?

theo25:
So far yes. If you could you review #106 that may help me add more tests.

theo25:
Jenkins: test this please

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -4132,6 +4139,22 @@ void IeleCompiler::appendMask(iele::IeleLocalVariable *Result, iele::IeleValue *
   }
 }
 
+iele::IeleLocalVariable *IeleCompiler::appendMemorySpill() {
```
dfilaretti: This can be reused in the `events` and encoding parts (and possibly others) if not done it already.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1035,12 +1035,10 @@ bool IeleCompiler::visit(const UnaryOperation &unaryOperation) {
       compileExpression(unaryOperation.subExpression());
     solAssert(SubExprValue, "IeleCompiler: Failed to compile operand.");
     // Compile as a subtraction from zero.
+    iele::IeleIntConstant *Zero = iele::IeleIntConstant::getZero(&Context);
```
dfilaretti: This is used quite a few times. It may be worth replacing other usages of `getZero(&Context)` with `Zero` (but `getZero(&Context)` might be useful too)

dfilaretti:
Looks OK as far as I can tell, couldn't see any obvious problem. 
Code is well commented and quite readable.

libsolidity/codegen/IeleCompiler.cpp
theo25: Yes I think it should be. If the encoding is merged before this I will do it.

libsolidity/codegen/IeleCompiler.cpp
theo25: I don't think there are other uses in this function. In general there may be places where that would make sense though, but I don't want to do this as part of this PR.

theo25:
Information about currently failing end to end tests with mappings:

- `mapping_state_inc_dec` : requires support for `uint256` as a fixed width type
- `structs`: requires support for recursive structs
- ` struct_reference` : requires support for recursive structs
- `array_accessor` : requires support for complex accessors
- `accessors_mapping_for_array`: requires support for complex accessors
- `complex_accessors` : requires support for complex accessors
- `struct_accessor` : requires support for complex accessors
- `bytes_inside_mappings` : fails because of using `msg.data`
- `struct_copy` : requires support for recursive structs
- `reusing_memory` : requires support for `keccak256`
- `accessor_involving_strings` : requires support for complex accessors
- `memory_types_initialisation` : requires support for encoding of non-scalar arguments to public functions
- `string_as_mapping_key` : requires support for `string` as mapping key
- `storage_string_as_mapping_key_without_variable` : requires support for `string` as mapping key
- `internal_types_in_library` : requires support for `string` as mapping key
- `using_library_structs` : requires support for `string` as mapping key
- `mapping_of_functions` : requires support for function pointers

libsolidity/ast/Types.cpp
```diff
@@ -2839,6 +2862,55 @@ string MappingType::canonicalName() const
...
+  // be too costly.
+  return true;
+  //return hasInfiniteKeyspace() || valueType()->isDynamicallyEncoded();
+}
+
+bigint MappingType::getFixedBitwidth() const {
+  if (hasInfiniteKeyspace()) {
+    return bigint(0);
+  } else if (hasHashedKeyspace()) {
+    return (bigint(1) << 256) * valueType()->getFixedBitwidth();
```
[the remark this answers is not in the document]
theo25: We resolved this offline. We agreed that the way hashed keyspace mappings are allocated makes having a unique prefix redundant.

PR #110, opened 2018-04-13
Minor fixes including to endianness

dwightguth:
Includes:

* a fix for copying storage to memory and copying to nested memory arrays
* a fix to a minor typo in a register name
* fix to make the tests pass now that iele is little endian in memory (still big endian in tx arguments)


theo25:
The new changes also seem fine to me.

Issue #17, opened 2018-03-12
Solidity feature: Account Call


theo25:
* [x] Value Types
* [ ] Reference types

PR #114, opened 2018-04-17
Encoding (revised)

### LINKED ISSUE #80: Encoding function to combine multiple integers into a single bytestring

dwightguth:
Contains code to encode and decode values to byte strings, and correctly invokes it when passing and returning reference types to external contract calls.

Fixes #80 #17 

The regressions come about because we now hit the missing case of encoding/decoding for function, bytes, and string types

Ready for review.


dwightguth:
Jenkins: test this please

PR #115, opened 2018-04-19
Separate types for (u)int and (u)int256

### LINKED ISSUE #109: separate uint and uint256, int and int256
Requires modifications to the frontend to make these separate types

theo25:
This PR adds `uint` and `int` as new arbitrary-width integer types. They are not syntactic sugar for `uint256` and `int256` as they used to be, and now the latter stand for the 256-bit integer types. The new types allow implicit conversions in the same spirit as the fixed bitwidth types, i.e. a conversion is implicitly allowed only from a type with shorter width to a type with larger.

This PR also adds a new list of command line tests to exclude (test/sould-skip-tests). These tests are completely skipped, i.e. we don't even attempt to compile them with solc. These are tests that are either extracted in error by our extraction script and are not actually solidity code or tests that are supposed to result in a compile error and are tested for this behavior in the C++ test suite, hence can be safely skipped from the command line tests.

Fixes #109 

This is ready for review.


theo25:
Jenkins: test this please

theo25:
Jenkins: test this please

theo25:
Jenkins: test this please

theo25:
Jenkins: test this please

iele-example-tests/auction/auction.iele.ref
```diff
@@ -66,6 +66,8 @@ entry:
   br %out.of.range1, throw
   sstore %_beneficiary_0, @"beneficiary"
   %timestamp = call @iele.timestamp()
+  %out.of.range2 = cmp lt %timestamp, 0
```
dwightguth: I'm not sure I like this. The timestamp can't ever actually be negative...

iele-example-tests/auction/auction.iele.ref
```diff
@@ -78,29 +80,35 @@ define public @"bid()"() {
...
   %tmp1 = iszero %tmp
   br %tmp1, throw
   %callvalue = call @iele.callvalue()
   %highestBid.val = sload @"highestBid"
-  %tmp2 = cmp gt %callvalue, %highestBid.val
-  %tmp3 = iszero %tmp2
-  br %tmp3, throw
-  %highestBidder.val = sload @"highestBidder"
-  %tmp4 = cmp ne %highestBidder.val, 0
+  %out.of.range3 = cmp lt %callvalue, 0
```
dwightguth: Same here, call value can't be negative

libsolidity/analysis/ReferencesResolver.cpp
```diff
@@ -214,8 +214,6 @@ void ReferencesResolver::endVisit(ArrayTypeName const& _typeName)
 		solAssert(!m_errorReporter.errors().empty(), "");
 		return;
 	}
-	if (baseType->storageBytes() == 0)
```
dwightguth: Why was this removed?

libsolidity/analysis/TypeChecker.cpp
```diff
@@ -1448,9 +1458,10 @@ void TypeChecker::endVisit(BinaryOperation const& _operation)
 			rightType->category() != Type::Category::RationalNumber
 		)
 			if ((
-				commonType->category() == Type::Category::Integer &&
-				dynamic_cast<IntegerType const&>(*commonType).numBits() != 256
-			) || (
+				commonType->category() == Type::Category::Integer && (
+				 dynamic_cast<IntegerType const&>(*commonType).isUnbound() ||
```
dwightguth: Shouldn't this be `!isUnbound() &&`?

libsolidity/ast/Types.cpp
```diff
@@ -1520,11 +1536,11 @@ MemberList::MemberMap ArrayType::nativeMembers(ContractDefinition const*) const
 	MemberList::MemberMap members;
 	if (!isString())
 	{
-		members.push_back({"length", make_shared<IntegerType>(256)});
+		members.push_back({"length", make_shared<IntegerType>()});
```
dwightguth: Shouldn't this stay the same, as a return type? Otherwise you won't be able to assign a length to a uint256 anymore.

libsolidity/ast/Types.h
```diff
@@ -347,12 +353,14 @@ class IntegerType: public Type
 	int numBits() const { return m_bits; }
```
dwightguth: Can we solAssert(!isUnbound()) here? I want to make sure we're not missing a case and somehow using the value 0 as if it's an actual number of bits.

iele-example-tests/auction/auction.iele.ref
theo25: Now that `uint256` is fixed width and since Solidity's `now` is typed as a `uint256` a conversion from `uint256` to `uint` has to take place before the add `now + _biddingTime`. I fixed the type conversion generation to not append a range check in this case.

iele-example-tests/auction/auction.iele.ref
theo25: This is due to the same problem above so it got fixed also.

libsolidity/ast/Types.cpp
theo25: I decided to make `length` a `uint`, since the array index is also a `uint` now. It seemed counter-intuitive to allow `uint` as the index value of an array index access, but restrict the length to `uint256`. However I am fine if you want to keep `length` as `uint256`. Let me know if you still think we should keep it that way.

dwightguth: Hmm I guess if the argument to new is a uint then it won't necessarily consume too much gas to allocate an array of >2^256 if that's the last allocation the contract ever performs, in which case we would lose type safety if we made length be a uint256. So I'm persuaded here. Maybe we could make length be of type uint256 for fixed length arrays though?

theo25: OK  I can do that.

libsolidity/analysis/TypeChecker.cpp
theo25: I agree the check is erroneous. I think it should be simply `!isUnbound()` though. What do you think?

theo25: We agreed that checking `!isUnbound()` is sufficient and this now fixed.

libsolidity/ast/Types.cpp
theo25: Done.

libsolidity/ast/Types.h
theo25: This is fixed now. I had to add a few extra guards (one that I missed before and two in places where `numBits` was called but its return value was ignored if the type was unbound).

libsolidity/analysis/ReferencesResolver.cpp
theo25: `storageBytes` is deprecated in the Solidity to IELE compiler. We haven't maintained its implementation and in general it is meaningless for IELE's storage. I removed because of this and because the way it's implemented right now it would return 0 for unbound integers. I also thought that there is no way that this returns other than unbound integers. I double checked that now, and I realized that the only other case it returns 0 is for a FixedBytesType with `m_bytes = 0`. I don't know what type this represents, since I am not aware of a `bytes0` type, but it is definitely allowed by the `FixedBytesType` constructor.

Do you have any ideas of how to proceed here? Also, please note that replacing this with `storageSize` is not feasible, since at the time of this call static storage allocation hasn't happened and `storageSize` of rerefence types like structs would return erroneous results.

theo25:
The 22 tests that are added to the should-skip list can be categoried as follows:
- 15 tests that are compile error and are checked in the c++ test suite for that behavior.
- 7 tests that are not solidity code, but json that get picked up by the extraction script.

The 8 tests that are added as regressions are all new hashes of tests that were failing before but their hash changed due to changes in their code.

Also I did not change any test that didn't include `(u)int256` or `(u)int` in its source code, if that's what you are asking.

PR #119, opened 2018-04-24
LValue refactoring

dwightguth:
Refactors out a class, IeleLValue, which manages reading and writing to lvalues.

The following LValue types exist:

* RegisterLValue: represents a local variable in IELE. Reading evaluates to the value in the register, writing evaluates to an assignment to the register.
* AddressLValue: represents a value stored in memory or storage. Reading performs sload or load, writing performs sstore or store.
* ByteArrayLValue: represents an element of a packed byte array: reading storage evaluates to an sload followed by a byte instruction, reading memory evaluates to a fixed width load, writing storage evaluates to an sload followed by an adjustment to the single byte followed by a sstore, and writing memory evaluates to a fixed width store.
* ArrayLengthLValue: represents the length member of an arbitrary width array. Reading evaluates to an sload or load, writing evaluates to instructions to erase the excess elements if the array is shrunk, followed by an sstore or store.
* ReadOnlyLValue: represents an lvalue that is not assignable directly. Used both to represent reads of values that might be an integer constant, as well as reads of reference types that are not represented via a pointer. Writing to such an lvalue requires a copy operation.


libsolidity/codegen/IeleLValue.cpp
```diff
@@ -40,7 +40,7 @@ ByteArrayLValue::ByteArrayLValue(IeleCompiler *Compiler, iele::IeleValue *Addres
 
 void ByteArrayLValue::write(iele::IeleValue *Value, iele::IeleBlock *InsertAtEnd) const {
   if (Loc == DataLocation::Storage) {
-    iele::IeleLocalVariable *OldValue = iele::IeleLocalVariable::Create(&Compiler->Context, "loaded.value", Compiler->CompilingFunction);
+    iele::IeleLocalVariable *OldValue = iele::IeleLocalVariable::Create(&Compiler->Context, "string.value", Compiler->CompilingFunction);
```
theo25: nit: shouldn't this be named `"string.val"` for uniformity?

libsolidity/codegen/IeleCompiler.h
```diff
@@ -195,32 +205,29 @@ class IeleCompiler : public ASTConstVisitor {
...
+      IeleLValue *To, const Type &ToType, iele::IeleValue *From, const Type &FromType, DataLocation ToLoc, DataLocation FromLoc);
 
   void appendAccessorFunction(const VariableDeclaration *stateVariable);
 
   void appendVariable(iele::IeleValue *Identifier, std::string name,
                       bool isValueType = true);
 
-  LValueKind loadKind(TypePointer type, DataLocation loc);
-  LValueKind storeKind(DataLocation loc);
+  IeleLValue *memberLValue(iele::IeleValue *Address, TypePointer memberType, DataLocation Loc);
```
theo25: I think this should have a better name given what it does. How about the more general `makeLValue`? Also rename parameter `memberType` to `type`.

PR #125, opened 2018-04-25
Fix CI and Regressions relating to strings

dwightguth:
Ready for review.


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -5017,10 +5021,18 @@ bool IeleCompiler::shouldCopyStorageToStorage(const IeleLValue *To,
...
+bool IeleCompiler::shouldCopyMemoryToStorage(const Type &ToType, const IeleLValue *To,
                                              const Type &From) const {
   return dynamic_cast<const ReadOnlyLValue *>(To) &&
-         From.dataStoredIn(DataLocation::Memory);
+         From.dataStoredIn(DataLocation::Memory) &&
+         ToType.dataStoredIn(DataLocation::Storage);
+}
+
+bool IeleCompiler::shouldCopyMemoryToMemory(const Type &ToType, const IeleLValue *To,
+                                             const Type &From) const {
```
dfilaretti: Having the two types named `ToType` and `From`, and the actual value named `To` can be a little confusing. Why not using e.g.:

```
bool IeleCompiler::shouldCopyMemoryToMemory(const Type &ToType, const IeleLValue *Value,
+                                             const Type &FromType) const {
```

theo25: If you do a renaming here, I would suggest only changing `From` to `FromType`. Also you should do it in both this function and the one below.

theo25:
The new changes also look good.

PR #126, opened 2018-04-26
Update to Solidity 0.4.23

dwightguth:
Ready for review.

**DO NOT MERGE**


test/libsolidity/syntaxTests/types/rational_number_bitshift_unlimited.sol
```diff
@@ -0,0 +1,8 @@
+contract c {
+    function f() public pure {
+        int a;
+        a = 1 << 4095; // shift is fine, but result too large
+        a = 1 << 4096; // too large
+        a = (1E1233) << 2; // too large
+    }
+}
```
theo25: Not sure I understand what the comments mean here.

dwightguth: This is a copy of rational_number_bitshift_limit.sol with type uint instead of type uint256. The comments are not mine.

theo25: OK maybe remove them? but I guess we have to do this kind of cleanup anyway at some point so I won't insist.

PR #129, opened 2018-04-26
ERC20 end-to-end tests

### LINKED ISSUE #130: ERC20 end to end tests

theo25:
This PR adds end to end tests for the ERC20 contract. The implementation of the contract is the same as found in `std/Token.sol` and `std/StandardToken.sol`. The tests are modeled after corresponding tests for the IELE implementation of ERC20, found [here](https://github.com/runtimeverification/iele-semantics/blob/master/iele-examples/erc20.iele). The corresponding IELE tests can be found [here](https://github.com/runtimeverification/iele-semantics/tree/master/tests/iele/ERC20). Tests related with no-overflow behavior have been skipped, since the Solidity ERC20 implementation we are testing uses fixed-width integer types, unlike the IELE one that works on unbound integers.

Fixes #130

### DIFF
--- a/test/contracts/ContractInterface.h
+++ b/test/contracts/ContractInterface.h
@@ -97,6 +97,41 @@ protected:
 		return h256(ret);
 	}
 
+	u256 callVoidReturnsUInt256(std::string const& _name)
+	{
+		bytes const& ret = callReturning(_name + "()");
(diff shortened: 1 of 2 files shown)


dwightguth:
Jenkins: test this please

Issue #122, opened 2018-04-24
Inline Arrays


dwightguth:
We need to support inline arrays.

Issue #121, opened 2018-04-24
Tuple desugaring


dwightguth:
Solidity has a syntax for nameless tuple components that we are not handling correctly everywhere, which. can lead to failed assertions and segfaults when building code with tuple expressions. We need to correctly support all these corner cases.

Issue #96, opened 2018-04-09
Recursive Structs


dwightguth:
Solidity supports recursive structs under certain circumstances. We need to compile these correctly.

Issue #61, opened 2018-03-26
Scoping Activation 


dfilaretti:
Seems we have a problem with variable scoping, as shown by test `C99_scoping_activation` (and possibly others, which will be added here as they are found), 

For example, consider a call to the following function:

```
function g() pure public returns (uint x) {
  x = 7;
  {
      x = 3;
      uint x;
      return x; // This should return the new variable, i.e. 0
  }
}
```

Instead, we return `3`. See `C99_scoping_activation` for more cases,

theo25:
According to documentation, this should return 0 (per C99 scoping rules) when compiled with solidity 0.5 and newer. We are still based on a 0.4 release, so that's why I implemented it to return 3 (per Solidity <0.5 scoping rules). Do they have tests that expect this to return 0 already? If that's the case, we indeed have a bug, as has their documentation.

dfilaretti:
Yes, the code above is copied from the test. 

The check is:
```
ABI_CHECK(callContractFunction("g()"), encodeArgs(0));
```

but it fails because we return 3.

PR #134, opened 2018-04-27
Miscellaneous bug fixes

dwightguth:
This fixes all the remaining tests in ABIEncoderTest, ABIDecoderTest, and SolidityEndToEndTest with the exception of the following categories of tests:

* function pointers (#49)
* C99 scoping (#61)
* named args (#22)
* recursive structs (#96)
* unsupported features (which are enumerated in README-IELE-SUPPORT.md)
* Tuple assignment (#121)
* Inline arrays (#122)

We also do the same for SolidityOptimizer with the exception that all of those tests still fail because we lack an optimizer still.

Finally, we fix the tests in StandardCompiler and JSONCompiler, with the exception that the gas estimates are still failing because we lack a gas analyzer still.

Note that many of the remaining tests fail due to lack of inline assembly, which these tests use in order to test other features of solidity that are not directly exposed by the solidity language. We will need to decide how to proceed with these tests as we may be missing test coverage of other features if we simply skip them all.

Ready for review.


theo25:
By the way, I agree with your comment related to the tests that include inline assembly. I don't think it's a good idea to blindly disregard them.

PR #138, opened 2018-04-30
Support for tuple assignment and inline arrays

### LINKED ISSUE #122: Inline Arrays

theo25:
This PR adds support for tuple assignments and inline arrays.

Fixes #122 #121 

This PR is ready for review.


theo25:
Jenkins: test this please

theo25:
Jenkins: test this please

theo25:
Note that commit 88e9c77 is the latest commit, that also passes CI, but github chooses to show commit 2cface2 as last in the list of commits. My best guess of why this happens is that commit 2cface2 has a more recent time stamp since it was committed before commit 88e9c77 but then reordered with it. I didn't know that github prioritizes time stamps.

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1294,64 +1294,235 @@ void IeleCompiler::appendConditional(
...
+        RHSValues.push_back(nullptr);
+        RHSTypes.push_back(nullptr);
+      }
 
-  RHSValue = appendTypeConversion(RHSValue, RHSType, LHSType);
+      solAssert(Components.size() == RHSValues.size(),
+                "IeleCompiler: failed to compile all elements of rhs of "
+                "tuple assignement");
+      for (unsigned i = 0; i < RHSValues.size(); ++i) {
+        if (const Identifier *Id =
```
dwightguth: Can't you swap something that isn't an identifier? Eg a.x, b.x = b.x, a.x

theo25: I will have to test that. My guess is that swap should work unless the values are in storage, so in this case it depends. In any case, it seems that the code here needs to be more general. I will fix this.

theo25: Actually, I am not sure it needs fixing. In the above example, a.x and b.x have both been evaluated and stored to temporaries by just visiting them. So swap will happen, I think. The only case this isn't true is when visiting identifiers that are local variables and that's the case taken care of here.

PR #140, opened 2018-05-01
RValue refactoring

dwightguth:
Refactors the compiler to create a IeleRValue class that represents an RValue of a single solidity value (which can occupy multiple IELE registers). We have limited testing for this functionality by means of function types, which are the only type which doesn't fit into a single register. Full support for function types will be present in the next PR.

Ready for review.


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -4118,32 +4182,32 @@ void IeleCompiler::endVisit(const Identifier &identifier) {
...
-        CompilingExpressionResult.push_back(AddressLValue::Create(this, Identifier, DataLocation::Storage, name));
+        return AddressLValue::Create(this, Identifier, DataLocation::Storage, name);
       } else {
-        CompilingExpressionResult.push_back(ReadOnlyLValue::Create(Identifier));
+        return ReadOnlyLValue::Create(IeleRValue::Create({Identifier}));
       }
-    } else if (auto Local = llvm::dyn_cast<iele::IeleLocalVariable>(Identifier)) {
-      CompilingExpressionResult.push_back(RegisterLValue::Create(Local));
+    } else if (llvm::isa<iele::IeleLocalVariable>(Identifier)) {
+      solAssert(false, "not implemented");
```
theo25: I was actually planning to remove this case and rename the method to `appendGlobalVariable`. Is this  a case we need to implement?

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -358,8 +358,8 @@ void IeleCompiler::compileContract(
 
           for (std::string paramName : paramNames) {
             if (!paramName.empty()) {
```
dfilaretti: Same here, if we decide to remove the check.

dfilaretti:
Looks good to me as well. I left a couple of minor comments.

PR #143, opened 2018-05-04
Recursive structs

### LINKED ISSUE #96: Recursive Structs

theo25:
This PR adds support for recursive structs.

- **Allocation**: Recursive structs in memory are no different than normal structs because when they allocated their dynamic array member that contains the recursion is allocated with length zero. Recursive structs in storage require extra care to not be infinite allocated. They are allocated lazily the first time they are accessed using a subclass of `ReadOnlyLValue`. This means that in presence of recursive structs we cannot know at compile time the size of the storage, hence we cannot support indexing of infinite keyspace mappings. For this reason, infinite keyspace mappings are demoted to hashed keyspace mappings in presence of recursive structs.

- **Copying/Deleting**: In order to avoid an infinite code generation when trying to generate code that deletes/copies recursive structs member by member,  we instead generate a recursive destructor/copier function for each recursive struct type.

- **Encoding/Decoding**: For similar to the above reasons, we generate recursive encoder/decoder functions for each recursive struct type. Note: that this functionality is never tested since Solidity doesn't currently support recursive structs as arguments or return values of public functions, members of events, or arguments to hash and log primitives.

Fixes #96 

This is ready for review.


libsolidity/ast/Types.cpp
```diff
@@ -1953,7 +1969,6 @@ unsigned StructType::calldataEncodedSize(bool _padded) const
 }
 
 bool StructType::isDynamicallyEncoded() const {
-  solAssert(!recursive(), "");
```
dwightguth: what stops this from recursing infinitely?

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -399,6 +439,27 @@ void IeleCompiler::compileContract(
     }
   }
 
+  // If we have to include the runtime storage, append code that initializes
+  // @ielert.storage.next.free global variable at the start of the constructor.
+  if (CompilingContract->getIncludeStorageRuntime()) {
```
dwightguth: shouldn't this be in appendDefaultConstructor instead?

libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -1687,6 +1758,7 @@ bool IeleCompiler::visit(const UnaryOperation &unaryOperation) {
     solAssert(LValues.size() == 1, "not implemented yet");
     TypePointer type = unaryOperation.subExpression().annotation().type;
     appendLValueDelete(LValues[0], type);
+    // TODO: What is the correct expression result value for delete? 0?
```
dwightguth: can you look this up in ExpressionCompiler and fix it rather than adding a todo?

libsolidity/codegen/IeleCompiler.cpp
theo25: appendDefaultConstructor is called multiple times and is called before any code generation happens for the body of any function. This should code should only be generated once in the actual constructor of the contract only if the storage runtime is needed, which can be known for certain after all code has been generated. That's why I added as the very last part of compileContract.

dwightguth: okay

libsolidity/ast/Types.cpp
theo25: The only potential way for it to recurse infinitely is because of a recursive struct. A recursive struct cannot contain itself directly it has to contain a dynamically sized array or mapping that uses the struct itself as the element type or the value type respectively at the very least. And if you look at the implementation of isDynamicallyEncoded for ArrayType (when the array is dynamically sized) and for MappingType, both of these return true without checking the element/value type hence the infinite recursion is broken.

dwightguth: Okay

libsolidity/codegen/IeleCompiler.cpp
theo25: I looked into ExpressionCompiler but I am not sure what to infer. It seems for example that deleting a local variable will leave the top of the stack as it was before starting to execute the delete. I am not sure how this translates in a non-stack based IR (or even it is intentional).

theo25: Given that, I think that not updating the CompilingExpressionResult vector seems like the actual best interpretation of the above behavior.

dwightguth: It seems that delete is of void type and therefore cannot be assigned to, which means this code was correct as is. If you remove this todo I will approve.

theo25: Removed!

PR #145, opened 2018-05-07
Major audit of all remaining non-passing tests

dwightguth:
The only tests after this is complete which we want to make pass but won't be passing yet should be the metadata tests that @theo25 is working on. Feel free to manually review the tests that still fail and let me know if you disagree with my assessment on any of them.

Ready for review.


dwightguth:
Jenkins: test this please

PR #149, opened 2018-05-09
 add errors for unsupported solidity features 

### LINKED ISSUE #142: Report error during analysis for unsupported features

dwightguth:
Fixes #142 

Ready for review.


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -4373,9 +4373,9 @@ bool IeleCompiler::visit(const MemberAccess &memberAccess) {
     solAssert(ExprValue,
               "IeleCompiler: failed to compile base expression for member "
               "access.");
-    solAssert(!(std::set<std::string>{"call", "callcode", "delegatecall"}).count(member),
-              "IeleCompiler: member not supported in IELE");
-    if (member == "transfer" || member == "send") {
+    if (member == "call" || member == "callcode" || member == "delegatecall") {
+    // cannot be invoked in iele, so we don't need to create a result.
```
theo25: Shouldn't this be caught by the syntax or type checker in the front-end and an error be reported? And if that's the case, I think an assertion here is fine.

dwightguth: I put the error on the function call expression instead, so this is actually now valid. You just can't ever call the function. That's why this code had to change.

PR #150, opened 2018-05-10
fix ipc with mantis on other end

dwightguth:
Do not review yet.

### DIFF
--- a/test/RPCSession.cpp
+++ b/test/RPCSession.cpp
@@ -129,7 +129,7 @@ string IPCSocket::sendRequest(string const& _req)
 		bytesread += ret;
 	}
 	while (
-		(ret == 0 || (bytesread > 0 && m_readBuf[bytesread - 1] != '}')) &&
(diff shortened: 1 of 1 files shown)


test/RPCSession.cpp
```diff
@@ -129,7 +129,7 @@ string IPCSocket::sendRequest(string const& _req)
 		bytesread += ret;
 	}
 	while (
-		(ret == 0 || (bytesread > 0 && m_readBuf[bytesread - 1] != '}')) &&
+		(ret == 0 || (bytesread > 0 && m_readBuf[bytesread - 1] != '}' && m_readBuf[bytesread - 1] != '\n')) &&
```
theo25: why do we need to check for both `{` and `\n`? Wouldn't just `\n` suffice?

dwightguth: well, the test harness we wrote doesn't emit a newline. I could change that I guess, but this seemed easier to me.

theo25: Souds good and I approved, but maybe we should consider the approach of having our test harness emitting newlines though

PR #153, opened 2018-05-17
Added documentation on how to avoid unsupported features

### LINKED ISSUE #31: Replacement for unimplemented features
We are not implementing several features of Solidity because they are EVM specific.  Usually they are also insecure, as well, and in a few cases they have been deprecated.   We need documentation that explains, for each feature, how a program with that feature should be implemented instead.

theo25:
This PR adds suggestions on how to avoid unsupported features in Solidity code when moving from EVM to IELE backend.

Fixes #31 

Ready for review.


dwightguth:
I'm going to merge this, but if @smalltalker55 wants further changes, he can let us know and we can make a followup PR.

PR #154, opened 2018-05-22
Cleanup

dfilaretti:
Goal of this PR:

* make clear that this is the Solidity to IELE compiler (READMEs etc.) 
* change the executable name (from `solc` to `isolc`) and update the command-line interface accordingly  
* possibly remove unneeded files, docs etc.


dfilaretti:
@dwightguth @theo25 I noticed several scripts and files I never used before and I wonder which ones are actually needed or supported and which ones we can get rid of. Some of them would need updating (since I changed the exacutable name from `solc` to `isolc`) but I'd rather not do it if they have to be removed. 

Here are some of the files: 

* `test/solcjsTests.sh`
* `snap/snapcraft.yaml`
* `scripts/travis-emscripten/publish_binary.sh`
* `scripts/release.sh`
* `scripts/release.bat`
* `scripts/release_ppa.sh`
* `scripts/Dockerfile`
* `scripts/docker_deploy.sh`
* `scripts/docker_deploy_manual.sh`
* `scripts/docker_build.sh`
* `scripts/bytecodecompare/storebytecode.sh`
* `scripts/bytecodecompare/storebytecode.bat`
* `scripts/build.sh`
* `README-IELE-INSTALL.md`
* `docs/using-the-compiler.rst`
* `docs/metadata.rst`

Additionally, the repository also contains some documentation in `docs/*`. Do we need to keep it or shall we just link to the "official" documentation?

There are also Travis and CircleCI files; since we use Jenkins, do we want to get rid of those?

dfilaretti:
I also notice the following warning from Jenkins output

```
+ gcovr -x -r .. -e ../test -o coverage.xml
+ kill %1 %2
Archiving artifacts
WARN: No artifacts found that match the file pattern "????-????-????-????". Configuration error?
WARN: ‘????-????-????-????’ doesn’t match anything
[Cobertura] Publishing Cobertura coverage report...
```
but I'm not exactly sure whether I introduced it or if it was there already.

dwightguth:
I can't answer which of these scripts might be important, since they were not created by us and presumably they have some reason why they were created by the EVM folks. My inclination would be to convert them all.

dwightguth:
@dfilaretti that warning can be ignored.

dfilaretti:
It looks like most of the extra scripts are for "internal use, and mostly have to do with Travis automation. 
Also, some of them deal with creating/releasing deb packages and uploading them to the ppa servers (e.g. `release_ppa.sh`) as well as handling Docker images. 

I tried to replace `solc` with `isolc` (our new executable name) where appropriate, but couldn't test all of those scripts and I don't think they apply to us, i.e. unless we use Travis or decide to publish deb packages, we may not need them right now (but I agree it may be useful to keep them just in case). 

Also notice for example that script `publish_binary.sh` contains something like 

```
git clone --depth 2 git@github.com:ethereum/solc-bin.git
cd solc-bin
git config user.name "travis"
git config user.email "chris@ethereum.org"
```
which obviously we need to update with our own data if we decide to use it.

README.md
```diff
@@ -1,5 +1,6 @@
-# The Solidity Contract-Oriented Programming Language
-[![Join the chat at https://gitter.im/ethereum/solidity](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/ethereum/solidity?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) [![Build Status](https://travis-ci.org/ethereum/solidity.svg?branch=develop)](https://travis-ci.org/ethereum/solidity)
+# The Solidity to IELE Compiler
+
+This is the Solidity to IELE compiler, a fork of the [Solidity compiler](https://github.com/ethereum/solidity) targeting [IELE](https://github.com/runtimeverification/iele-semantics) instead of the EVM.
```
theo25: I think you should also update the the `Building` section with a pointer to the instructions in `README_IELE_INSTALL.md`. Also remove the `How to Contribute` section as it refers on how to contribute to the EVM compiler.

theo25:
It looks good to me, given that the changes on the scripts (or at least the ones that we are interested in) have been tested. Also I am wondering whether we want to rename `soltest` to `isoltest` as well. @dwightguth ?

dwightguth:
Don't rename soltest to isoltest because ironically, there is already an isoltest in solidity-to-evm. I'm operating on the assumption however that as a testing binary, soltest will not be installed on users machines so it doesn't need to be renamed.

README.md
dfilaretti: Done!

dfilaretti:
@theo25 @dwightguth  I tested the main scripts such as `ciscript`, `scripts/build.sh` (although we never actually use it), `scripts/release.sh` (which does a build and then makes a zip file with the executables), `ieleCmdLineTests.sh` (used by `ciscript`) etc. 

I could not test the Travis automation scripts, the Docker ones (which they use to build and upload https://hub.docker.com/r/ethereum/solc/) and the deb packages & ppa stuff.

dfilaretti:
Jenkins: test this please.

dfilaretti:
I merged the various README into one, which now contains basic info as well as build instructions, so that there's a single entry point, `README.md`

PR #156, opened 2018-06-04
remove warning related to unsupported feature

dwightguth:
The warning is redundant now that we have an error that rejects programs that use this feature.


dwightguth:
@theo25 this is ready for review.

PR #157, opened 2018-06-05
Fix for llvm 6

dwightguth:
Several bugs were discovered in testing today related to llvm 6. This PR should now work on both LLVM 5 and 6.


theo25:
@dwightguth should I review this or do you plan to add more changes?

theo25:
OK it seems the reason for failing is that #156 needs to be merged first. In this case I will approve the changes.

PR #162, opened 2018-06-27
Fix encoding of negative arbitrary-precision integers

### LINKED ISSUE #161: Fix encoding of negative arbitrary-precision integers
There is a bug in the code generated for encoding of negative arbitrary-precision integers where the generated code attempts to compute the logarithm of a negative number.

theo25:
Fixes #161

### DIFF
--- a/libsolidity/codegen/IeleCompiler.cpp
+++ b/libsolidity/codegen/IeleCompiler.cpp
@@ -2143,45 +2143,61 @@ void IeleCompiler::appendByteWidth(iele::IeleLocalVariable *Result, iele::IeleVa
    // width_bytes(n) = ((log2(n) + 2) + 7) / 8
    //                = (log2(n) + 9) / 8
    //                = (log2(n) + 9) >> 3
+
+   // Check for positive value.
+   iele::IeleLocalVariable *IsPositive =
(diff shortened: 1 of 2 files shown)


libsolidity/codegen/IeleCompiler.cpp
```diff
@@ -2143,12 +2143,41 @@ void IeleCompiler::appendByteWidth(iele::IeleLocalVariable *Result, iele::IeleVa
...
+      appendConditional(IsPositiveValue, PositiveResults,
+       [&Value, this](llvm::SmallVectorImpl<IeleRValue *> &Results){
+        Results.push_back(IeleRValue::Create({Value}));
+       },
+       [&Value, this](llvm::SmallVectorImpl<IeleRValue *> &Results){
+         iele::IeleLocalVariable *PositiveResult =
+           iele::IeleLocalVariable::Create(
+             &Context, "positive.value", CompilingFunction);
+        iele::IeleInstruction::CreateBinOp(
+          iele::IeleInstruction::Sub, PositiveResult,
```
dwightguth: Pretty sure this should be a bitwise not. You could also adjust for twos complement by replacing 0 with -1, but not showing use less gas.

PR #163, opened 2018-06-28
Add some initial help pages

marick:
First examples of help for people who are moving from EVM-Solidity to IELE-Solidity.


dwightguth:
Jenkins: add to whitelist

help/README.md
```diff
@@ -0,0 +1,137 @@
...
+  IELE-Solidity.
+
+* **Check uses of `ecrecover`**
+  
+  The EVM-Solidity version of `ecrecover` returns `0` in case of
+  error. Code that doesn't check for the error case might incorrectly
+  and irreveribly send funds to address `0`.
+  
+  Therefore, the IELE-Solidity version throws an exception instead. So
+  if you *did* check the return value, you can now remove that check.
```
theo25: That is not exactly correct. One still needs to check the return value (or more accurately the status return value) in order to figure out whether the call was successful or threw an exception.

help/call.md
```diff
@@ -0,0 +1,71 @@
...
+    store.call(callee, 1, 2, 3);
+
+Because `call` is a security risk, IELE-Solidity doesn't support
+it. Instead, `set` is called just like any other public function:
+
+    store.set(...);
+    
+    
+`store.set` takes a single argument (of type `bytes`) that must
+conform to the ABI.
```
theo25: This statement may be misleading. There is no ABI for IELE. And even in EVM, using the low-level call means that you send any bytestream you want, it doesn't need to conform to the standard EVM ABI. The only thing needed is a common understanding between caller and callee on how to encode/decode the bytestream.

That being said, if you use the low-level call of EVM-Solidity with arguments of types other than `bytes` (as in the example, where `uint256` is used), then the compiler will pack the arguments to a bytestream according to the EVM ABI. That's why the IELE-Solidity code uses `abi.encode` to maintain equivalent functionality with the EVM-Solidity code.

help/call.md
```diff
@@ -0,0 +1,71 @@
...
+
+The target function's name is encoded the same way as in EVM-Solidity:
+
+    bytes4 callee = bytes4(keccak256("set()"));
```
theo25: Similalrly here: It should be made clear that we do all this in the IELE-Solidity code, just to generate the exact same bytestream that the low-level call will generate and send in the EVM-Solidity code.

The way this section is written right now it makes seem like one has to encode the function's name in IELE-Solidity in order to call that function. This isn't true. In IELE-Solidity, the simple `store.set(..)` with whatever bytestream data as argument will call the `set` function of the callee contract. You don't have to encode the bytestream according to the ABI and you don't have to include the encoded function name. We do all these things here simply to send the exact same bytestream that EVM-Solidity would have sent.

help/README.md
marick: I don't understand that in the light of this from https://github.com/runtimeverification/solidity/blob/sol2iele/README-IELE-SUPPORT.md 

screen shot 2018-07-18 at 5 38 54 pm

I can't find any use of `ecrecover` that *does* check the return value, but I imagine it would look like this:

```
addr = ecrecover(...);
if (addr == 0) { 
  ... handle error ...
} else { 
  ... handle success
}
```

How would that be changed?

help/assembly.md
```diff
@@ -0,0 +1,49 @@
+# Error: Inline assembly is not supported in IELE.
+
+IELE-Solidity doesn't allow embedded assembly code. Many uses of assembly
+have Solidity-level replacements.
+
+### Elliptic curve functions
+
+The `ECADD`, `ECMUL`, and `ECPAIRING` bytecodes have corresponding
```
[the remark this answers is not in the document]
marick: Done

help/assembly.md
```diff
@@ -0,0 +1,49 @@
...
+     
+     uint16 size = addr.codesize;
+      
+
+Note that `codesize` will still return `0` if no code has been deployed to the given address.
+
+* Before: [examples/codesize.evm.sol](examples/codesize.evm.sol)
+* After: [examples/codesize.iele.sol](examples/codesize.iele.sol)
+
+### Reviewers: Anything else?
```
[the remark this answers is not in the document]
marick: Done

help/call.md
marick: Perhaps it would be better to have an example that does more with the `bytes` argument than assign it to a variable. I don't know enough to extend this example that way.

marick: @theo25 Hate to be a pest, but IOHK is drawing from my examples, and their doc freeze is Monday.

help/README.md
theo25: Yes, you are right. My mistake, you don't need to check in the Solidity level, which you are discussing here. So this text is fine, no need to change it.

help/call.md
theo25: I don't think an example that does more with the argument would help here, since here we are discussing the call-site code, not the callee code.

What you need to say is that in the EVM-Solidity code, the arguments to `store.call(...)` are encoded according to the EVM ABI into a single bytestring which then is passed to the callee. Moreover the callee is identified as the function `set` by decoding the incoming bytestring at the callee contract.
In IELE-Solidity on the other hand we do not support this type of low-level behavior. One needs to do an account call normally, as in `store.set(..)`. If we want our IELE-Solidity program to pass to the callee function the exact same bytestring as in the EVM-Solidity case, we have to manually construct it. In this case we do as such: `abi.encode(bytes4(keccak256("set()")), uint256(1), uint256(2), uint256(3))`.

And then you can proceed explaining the details of why we encode this way as you have done already

marick: Done, I hope.

help/call.md
```diff
@@ -0,0 +1,71 @@
...
+non-overflowing `uint` type. Because they're no longer the same type,
+IELE-Solidity must encode `uint` arguments differently than `uint256`.
+In the above example, they'd be encoded differently than an
+EVM-Solidity compiler would.
+
+You might not care. Perhaps you haven't written `ByteStoreContract` yet, so
+you're fine with arbitrarily large integers. 
+If you're working with a pre-existing contract, though, take care not to send
+`uint` values to a function that expects `uint256`. In such a case, explicitly cast
+the arguments:
```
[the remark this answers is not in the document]
marick: Done, I hope.

help/call.md
marick: Done, I hope

help/call.md
```diff
@@ -0,0 +1,77 @@
...
+it. 
+
+You can emulate what `call` does by constructing a bytestring that's
+byte-for-byte identical to the one `call` would make. Here's a first
+attempt:
+
+    bytes4 callee = bytes4(keccak256("set()")); -- same as above
+    abi.encode(callee, 1, 2, 3)
+
+The reason the result isn't identical is that, in IELE, `uint` and `uint256`
```
theo25: IELE -> IELE-Solidity

dwightguth:
Jenkins: test this please

PR #175, opened 2021-03-29
Add Nix expression

ttuegel:
Fixes: runtimeverification/iog-pm#64

This is my work-in-progress packaging `isolc` with Nix. Actually, I think the package is done, but I need to add the test suite and connect everything for continuous integration.


ehildenb:
Is this something that the upstream solidity repo could take advantage of? If so, we may consider getting it merged there, so that users of the normal solidity compiler can take advantage of it.

ehildenb:
Sounds like no, because they already have that taken care of.

theo25 (approved):

ehildenb:
Is this tested on CI currently? Should we add github actions for it @ttuegel ?

ttuegel:
I have disabled the GitHub Actions on macOS temporarily, until we resolve runtimeverification/iele-semantics#268.

ttuegel:
As it turns out, retrieving items from the cache works better when you point it at the right cache! :sweat_smile:

PR #178, opened 2021-04-05
Unchecked

mariaKt:
This PR implements:
- the `unchecked{...}` feature and checked arithmetic in its absence.
- the new syntax for the gas and value function call options.


theo25 (approved):
Looks good to me.

PR #190, opened 2021-04-22
Milestone 4 Regression Tests

gtrepta:
fixes: https://github.com/runtimeverification/iog-pm/issues/75


Dockerfile
```diff
@@ -11,7 +11,8 @@ RUN    apt update        \
        llvm-11-dev       \
        libz3-dev         \
        libxml2-utils     \
-       gcovr
+       gcovr             \
+       npm
```
ehildenb, line 15: Let's keep these sorted.

Jenkinsfile
```diff
@@ -25,6 +26,28 @@ pipeline {
...
+            FAILING = "${env.WORKSPACE}/test/milestone4-contracts.failing"
+          }
+          steps {
+            dir('iog-pm') {
+              sh '''
+                git submodule init
+                git submodule update --depth 1
+                grep -v -x -F -f ${FAILING} ${CONTRACTLIST} > contracts.txt
+                export SOLC=${WORKSPACE}/build/solc/isolc
+                ./run-all-scripts.sh --filter contracts.txt --verbose --verbose
```
theo25, line 44: Is the double `--verbose` flag needed here or just a typo?

theo25 (approved):
This looks good to me. I just left a minor comment/question.

Jenkinsfile
gtrepta: It's needed. The first option is used by `run-all-scripts.sh`, and the second one gets passed to `repos/run-solc.sh` which prints out warnings/errors from isolc. The harness is hacked together so it isn't elegant at this stage.

PR #191, opened 2021-04-23
Nested tuples

mariaKt:
This PR adds support for nested tuples.

We now allow returning more than one expression/lvalue when visiting a tuple.

Also, we change the way assignment works. We initially push the types/expressions of the tuple in a vector and reverse them. Then, starting from the end of the vector, traverse the types/expressions. If not a tuple, we push the found element in the final element list. If a tuple, we remove this element and push all its elements (types/expressions) in the vector, again in reverse order.
This order is needed because that is how the nested tuple elements are returned.

Solves https://github.com/runtimeverification/iog-pm/issues/86


theo25 (approved):
Looks good to me!

PR #196, opened 2021-04-29
Add macOS to Test workflow

ttuegel:
This is possible now that iele-semantics is built on macOS.


ttuegel:
The vanilla Solidity compiler does not build on macOS with Nix, so we have no hope of building `isolc` either.

PR #275, opened 2021-09-01
Adding a new abi field, ieleName, with the name of the iele function …

theo25:
This PR adds a new ABI field named `ieleName` that contains the IELE generated name that corresponds to each Solidity function.


theo25:
I am closing this since it turns out the current way of outputting the ABI for function names is fine.

